PR #763, opened 2015-01-08
Corrected escaping in Trac regex

psiinon:


ElbertF:
You shouldn't need to escape slashes. Does the current pattern not work?

psiinon:
My bad - there was a bug in my code and I thought this was the problem. It wasnt :)

PR #767, opened 2015-01-12
Added Phusion Passenger

dvrylc:
Identifies the [Passenger web/app server](phusionpassenger.com) instead of just Ruby on Rails *(which runs on other servers like Unicorn as well)*


ElbertF:
Thanks!

PR #776, opened 2015-01-29
Update the apps.json path

nicolas-talichet:


ElbertF:
Thanks, but this isn't right. `apps.json` Is supposed to be in the `drivers/php` folder, running `links.sh` or `links.cmd` will copy it there.

nicolas-talichet:
Oh ok I see, thanks for the explication and have a good day.

PR #778, opened 2015-01-29
Elm language / JS framework added

toastal:
[Elm](http://elm-lang.org/). I'm not 100% sure if this should be considered a JS framework or not. It's a Haskell-y-looking functional reactive language that compiles to essentially a JS framework.

I used the code from Ember.js that just looks for Ember in the env as a guide.

I could consider searching scripts for `elm?.+\.js`, since the base compilation of an `.elm` file using `elm-make` is a file called `elm.js` and there's also the base runtime on [CDNJS](https://cdnjs.com/libraries/elm-runtime) as `elm-runtime.min.js`, but is it necessary/useful/accurate?

Can test @ 

Elm - http://elm-lang.org/


ElbertF:
Thanks! Looking for JS files in the HTML source would be useful for drivers that don't evaluate JavaScript, such as the PHP driver.

PR #780, opened 2015-01-30
Added Cowboy

toastal:
Cowboy doesn't list a version number it seems. I also trimmed that whitespace on mentioned in my proposal for EditorConfig.


ElbertF:
Thanks!

PR #781, opened 2015-01-30
Added RxJS

toastal:
RxJS is a functional reactive library from Microsoft.RxJS seems to have a naming convention for its flavors: https://cdnjs.com/libraries/rxjs . I added the `env` namely to cover [Netflix](https://www.netflix.com/us/) whose RequireJS will show no script tag, but they've make it very clear that they love the FRP Rx provides them.


ElbertF:
Thank you!

q--:
The environment variable is too short to have 100% confidence when matching. Such short variables can exist in compiled source code such as that generated by the Google Closure Compiler.

For example, I can see a function named Rx at docs.google.com.

toastal:
Hmm.. If I recall correctly, the `env` match can't match any objects deeper than what's at the global scope for instance `Rx.Observable`. And I suppose there's no way to rule out the use of Closure compiler/Uglify w/ mangling/etc. What's done in these cases? Is the confidence set lower? Is it removed? I mean to get to 'Rx' you would have had to have set and not reused over 1100 global variables, no? Google Docs takes all env objects `[A-Za-z0-9_$]{1,3}`, correct? I think most utility libraries try to keep the name short ala `$()`.

q--:
@toastal Google Docs does indeed have (global) variables with these characters.

toastal:
So the long and the short of it is that no environment variable check for a string of 1 to 3 length is safe because of Google Docs has reserved them all? Seems like that that should be in the [contributing doc](https://github.com/ElbertF/Wappalyzer/wiki/Contributing).

q--:
I added something to the Wiki: https://github.com/ElbertF/Wappalyzer/wiki/Specification#hints--pitfalls

By the way, @ElbertF, was the confidence threshold ever implemented? (E.g. that applications with a confidence lower than *X* are hidden from the list.) I'm asking this because I see that RxJS adds two other (quite generic-sounding) environment variables which could used to reduce false positives.

ElbertF:
No, that hasn't been implemented yet.

PR #783, opened 2015-02-09
Lower RxJS env pattern confidence

q--:
See the discussion here: https://github.com/ElbertF/Wappalyzer/pull/781#issuecomment-72619339


ElbertF:
Thanks!

PR #787, opened 2015-02-12
again fix duplicate `html` property in `Trac` app

totpero:
In: `Trac` app is two property with the same name `html` one is string and one is array of strings:
            "html": "<a id=\"tracpowered",
            "html": [ "<a id=\"tracpowered", "Powered by <a href=\"[^\"]*\"><strong>Trac(?:[ /]([\\d.]+))?\\;version:\\1" ],


ElbertF:
Thank you!

PR #788, opened 2015-02-13
Added SquirrelMail

q--:
Another webmail client. Not much to detect because the main part of the application uses frames, which are not scanned by Wappalyzer. Still, the URL should be sufficiently distinctive.


ElbertF:
Can't believe I hadn't already added this one, thanks. :)

PR #789, opened 2015-02-13
Added Woocommerce Quick Donation

technofreaky:
Added Woocommerce Quick Donation
https://wordpress.org/plugins/woocommerce-quick-donation/


ElbertF:
Thanks for contributing! With a few exceptions (Woocommerce itself for instance) I prefer not to include plugins, considering there are so many and the most important information is the platform itself.

PR #791, opened 2015-02-17
Fixed regular expressions broken in commit 5a015f2

Datacast:
Looks like commit 5a015f2 broken apps.json processing (Parse error).  I made this change and confirmed it fixed the issue.

### DIFF
--- a/drivers/chrome/background.html
+++ b/drivers/chrome/background.html
@@ -4,6 +4,7 @@
 	<head>
 		<title>Background</title>
 
+    <script src="js/spotkick-extension-utils.js"></script>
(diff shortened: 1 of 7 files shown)


ElbertF:
I fixed this locally but it looks like I forgot to push my changes. Your pull requests includes other files, probably not what you intended.

Datacast:
Thanks- that's odd.  When I did the pull it clearly only had apps.json
affected.  Immediately after I tweaked files so I wonder if that's what
happened.

I'll make it cleaner next time.  Thanks

Eric

PR #792, opened 2015-02-19
fix missing mod_dav icon

totpero:


totpero:
test here: http://webdav.org/mod_dav/

ElbertF:
Thanks!

PR #796, opened 2015-02-23
add OWL Carousel

NovikovViktor:


src/apps.json
```diff
@@ -7005,7 +7005,19 @@
...
+		},
+        "OWL Carousel": {
+            "cats": [
+                5,
+                7
+            ],
+            "html": "<link [^>]*href=\"[^\"]+owl.carousel(?:\\.min)?\\.css",
+            "script": "owl.carousel.*\\.js",
+            "implies": [
+                "CSS", "jQuery"
```
ElbertF: There is no application called `CSS`, please change it to `"implies": "jQuery",`

DaAwesomeP: This is also not in alphabetical order if that matters.

ElbertF: That's ok, running `wappalyzer build` will fix the order and indentation.

NovikovViktor:
@ElbertF , I change `"implies"` part, so it can be merged...

ElbertF:
Thanks!

PR #798, opened 2015-02-23
Add simple JSON validator Travis CI Test

DaAwesomeP:
What this does:

 - Running `npm test` (like Travis CI does) will check for validation on `apps.json` and `schema.json`.
 - I am using Grunt (an NPM package) to test it.
 - More tests can be added; this is just a basic one to catch the most common errors. Most pull requests are to `apps.json`.
 - Grunt can also be used for automated building, etc.
 - All pull requests will automatically be tested and their pass/fail will be shown in the pull request,

What will need to be done if this is merged:

 - This repo will need to be registered on Travis CI (free). I already have added the `.travis.yml` file, so it should start working immediately.
 - A badge should be added to the top of the README. The badge won't work until Wappalyzer is registered on Travis CI. It is the passing/failing badge: `[Travis]()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@
 
 build/*
 
+npm-debug.log
+
 node_modules/*
 
 drivers/npm/node_modules/
+
+drivers/npm/npm-debug.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+﻿language: node_js
+node_js:
+  - "0.10"
+before_install:
+  - export WAPPALYZER_ROOT=$TRAVIS_BUILD_DIR
+  - export PATH=$PATH:$TRAVIS_BUILD_DIR/bin
+install:
+  - sudo apt-get update -y
+  - sudo apt-get install -y curl zip sudo -y --force-yes
+  - sudo apt-get clean
+  - npm install jsonlint -g
+  - mkdir mozilla && curl -L https://ftp.mozilla.org/pub/mozilla.org/labs/jetpack/jetpack-sdk-latest.tar.gz | tar xvzC mozilla && ln -s $WAPPALYZER_ROOT/mozilla/addon-sdk-*/bin/cfx bin/cfx
+  - mkdir phantomjs && curl -L https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-linux-x86_64.tar.bz2 | tar xvjC phantomjs && ln -s $WAPPALYZER_ROOT/phantomjs/phantomjs-*-linux-x86_64/bin/phantomjs bin/phantomjs
+before_script: mkdir build
+script: wappalyzer build
+after_script: ls build
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Wappalyzer
+# Wappalyzer [![Travis](https://img.shields.io/travis/ElbertF/Wappalyzer.svg?style=flat-square)]()
 
 [Wappalyzer](https://wappalyzer.com/) is a 
 [cross-platform](https://github.com/ElbertF/Wappalyzer/wiki/Drivers) utility that uncovers the 


ElbertF:
Great, thanks!

How do you feel about using Gulp instead of Grunt? The config is cleaner and is supposed to be faster.

I'm planning to have `wappalyzer validate` do much more than just validate the JSON against the schema. This means tests might pass on Travis but not pass validation. I guess Travis still adds some value but we're basically testing in two places.

DaAwesomeP:
I've never used Gulp before, but I've always needed to. I'll try to get it working with that and add another commit to remove Grunt and add Gulp.

The great thing about Grunt and Gulp is that they can do just about everything. It can be expanded to test whatever needs to be tested. The JSON is where most of the errors come from, and it's also the easiest thing to test, so I went ahead and added it.

ElbertF:
I'm a fan of task runners but opted for shell scripts in this case ([see here](https://github.com/ElbertF/Wappalyzer/blob/master/bin/wappalyzer-build)), they're a bit easier to do more complicated things with and have fewer dependencies. They're not very cross-platform but they don't need to be.

One of the things I would like to test is that the patterns are all valid regex, test the Wappalyzer specific fields (`version` and `confidence`), make sure the HTML patterns don't match plain text, etc.

DaAwesomeP:
I suppose you could run the `wappalyzer` commands through Travis. I'm not sure that they would properly trigger pass/fail, but it's worth a shot.

I'm going to commit and try some things in my fork. I think I can get away with using the scripts in `bin/`.

DaAwesomeP:
@ElbertF, there you go. It is validating with `./bin/wappalyzer-validate`. I think that we should add `npm install -g jsonlint` to `wappalyzer-validate` so that if more dependencies are needed, then `.travis.yml` won't have to be updated too. NPM will only install the package anyways if it is out of date or doesn't exist.

DaAwesomeP:
Once you want to add regex testing and the other stuff, you will definitely need Gulp. I'm not sure its worth the trouble of parsing JSON in a shell script.

gadcam:
@ElbertF @DaAwesomeP You can validate a field of the JSON with a regular expression with JSON schema, I think the better it to keep the greatest part of the validation in the JSON schema, if we can not do it all in. (See https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.16)

DaAwesomeP:
I see. So the JSON validation covers everything.

We still need to include a script for the dependencies or simplt add them to the `wappalyzer-validate` script. I am manually including `npm install -g jsonlint` in `.travis.yml`. @ElbertF are you okay with me adding that to `wappalyzer-validate`?

ElbertF:
@DaAwesomeP No, it shouldn't go in `wappalyzer-validate`. It's already in the Dockerfile, making jsonlint globally available in the development environment. If Travis needs it I think an include script is fine.

ElbertF:
> Once you want to add regex testing and the other stuff, you will definitely need Gulp. I'm not sure its worth the trouble of parsing JSON in a shell script.

I don't think we'd need Gulp, the shell script could run a JavaScript file through node to do the validation. I wouldn't use JSON schema or regex for this.

DaAwesomeP:
Don't close this yet; I'm trying one last thing in one of my VMs. I am going to see if Travis CI can run Vagrant. Then we can automatically test and build everything.

ElbertF:
Ha, that would be amazing. It's a lot of overhead though.

DaAwesomeP:
I hope this works...

ElbertF:
Try `apt-get install docker -y` instead of `apt-get install docker.io -y`.

DaAwesomeP:
In the docker docs it says to use docker.io. They are different packages. I think this is an older version of Ubuntu (possibly 12.04). I'm just going to use the more up-to-date manual way that they list.

https://docs.docker.com/installation/ubuntulinux/

DaAwesomeP:
It stalled doing the exact same thing is did before. I'll try re-adding the `apt-get update`.

DaAwesomeP:
It stalled again. I'm stumped. It could possibly be a particular Travis CI server, but that is unlikely.

DaAwesomeP:
I've sent a support email to Travis CI about the inconsistent stalls. Sometimes it stalls and sometimes it doesn't. I don't think it is a problem with `.travis.yml` or anything else that we can control.

DaAwesomeP:
I think that the solution would be not to mess with Docker or Vagrant in Travis at all. Unless there is a way to run a dockerfile like you would a sh, we need to simply have the dockerfile run an sh with all of the setup commands. This way Docker and Travis would run the same setup and build scripts without 1) having a VM inside a VM in Travis. 2) Having a Dockerfile and sh script for setup that must both be updated at the same time and match. 3) Save any more hassle with Travis.

@ElbertF are you okay with this? I could very easily convert most of te dockerfile into an sh. I will test it in my VM before I commit it.

ElbertF:
I agree that setting up Vagrant & Docker on every commit is probably a bit much. I wrote the shell scripts in a way that they can be run from outside the Docker environment, as long as the dependencies are installed. I don't mind having a setup script in two places, one for Docker and one for Travis.

DaAwesomeP:
Ok, I'll make an `install-dependencies.sh` in the root. I won't touch the dockerfile. This means though that if the dockerfile is changed, then the other file will need to be changed too.

ElbertF:
Great. Should we perhaps keep it in `.travis.yml`? It's not a script anyone should run.

DaAwesomeP:
That's a good idea. I'll do that.

DaAwesomeP:
@ElbertF It is done! Here is a passed test: https://travis-ci.org/ElbertF/Wappalyzer/builds/52068803#L1577

However, when parts failed they did not trigger Travis. This is a problem in `wappalyzer-build`. The build sequence needs to (a little uncleanly) quit completely with an error if there is a problem. In one build (https://travis-ci.org/ElbertF/Wappalyzer/builds/52068059#L1576), Travis passed, but the build really failed.

I will fix that and use that commit to squash all of the others. I'm not sure how to fix that though (i.e. how to make the script exit on error).

DaAwesomeP:
Actually, from looking at `wappalyzer-validate`, it has code to exit correctly. The problem is only in `wappalyzer-build`. Should I add the following under each build command for Firefox, Chrome, and the Bookmarklet?
```bash
if [ $? -eq 0 ]
then
	echo -e "${color_ok}Ok${color_none}"

	exit 0
else
	echo -en "${color_none}"

	exit 1
fi
```

ElbertF:
I'll have a look at the shell scripts, there may be a more elegant way to handle errors.

DaAwesomeP:
[Here](http://www.davidpashley.com/articles/writing-robust-shell-scripts/#id2382181) lists several ways to do it. For now it should catch the `apps.json` errors (the most likely ones), so I will add the badge to the README and squash the commits. Then it will be ready to merge.

DaAwesomeP:
There we go! One commit with the `.travis.yml` and the badge in the README.

ElbertF:
Excellent, thanks!

AliasIO:
`set -e` seems elegant enough to catch errors. Done in ea1f3553cb644181761fd6fc262fb4a76d011c37

PR #801, opened 2015-02-26
Wappalyzer NPM Driver fix package.json typo, updated tests, added jshint, and fix multiple req

pasindud:


AliasIO:
Thanks!

Do we need another .editorconfig? It's the same as the one in the root.

pasindud:
oh yea no need that one both are the same, made the wappalyzer cli also with nodejs https://github.com/pasindud/wappalyzer-cli will send you the change with that :+1:

PR #802, opened 2015-02-26
Added Lithium signature and icon

atcazzual:
Added Lithium signature and icon


AliasIO:
Thanks! Are you able to use a 32x32 icon?

gadcam:
The HTML pattern is too broad I think.
You can find a 32x32 icon at http://community.lithium.com/html/assets/stars-icon-li.jpg.

AliasIO:
Ah, you're right. The HTML pattern should have at least an HTML tag in it.

atcazzual:
Yeah, I can grab the 32x32 icon.  Do we prefer it to have a transparent background?

As for the HTML pattern would 
    alt="Powered by Lithium"
be ok?  I would rather avoid having the pattern too strict like matching exactly on the entire anchor tag as if Lithium changes it we will no longer be detecting it.

atcazzual:
Ok, so to get an HTML tag in the pattern, how about the following?
    <a\s(.(?!</a>))*alt="Powered by Lithium"(.(?!</a>))*.</a>

And do we prefer a transparent background on the icon?

atcazzual:
ok, so that pattern didn't show up right.  How about now?
    &lt;a\s(.(?!&lt;/a&gt;))*alt="Powered by Lithium"(.(?!&lt;/a&gt;))*.&lt;/a&gt;

AliasIO:
Transparent background, yes.

Do you have an example website where that string appears?

gadcam:
You can also look for these cookies
<code>
LithiumVisitor
LiSESSIONID
LithiumVisitor
</code>
and this DOM variable <code>LITHIUM</code>.

atcazzual:
Example link: http://community.starhub.com/t5/Broadband-Recontract-Offers/urgent-require-a-reply-from-starthub-boardban-service/td-p/86977

Also just updated icon to 32x32, updated HTML pattern, added "LithiumVisitor" cookie to "headers" attribute and added "LITHIUM" to the "env" attribute.

AliasIO:
The HTML pattern can be simplified: `<a[^>]+Powered by Lithium`

atcazzual:
Should we at least keep the \s in so that it does not match on any of the following other html tags:
&lt;abbr&gt;, &lt;acronym&gt;, &lt;address&gt;, &lt;applet&gt;, &lt;area&gt;, &lt;article&gt;, &lt;aside&gt;, &lt;audio&gt;

AliasIO:
Fair enough, you can just add a space: `<a [^>]+Powered by Lithium`.

atcazzual:
Done.

klonos:
@atcazzual ...

> ok, so that pattern didn't show up right.

You can wrap code in "```" (or just add four spaces at the beginning of a line) and the GitHub markdown won't mess with what you type ([Fenced code blocks](https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks)).

atcazzual:
Is there any other suggestions on this signature?  Or is it good?

I ask because there is another signature that I would like to add but do not want to combine it in the same pull request.

gadcam:
@atcazzual : It seems good to me ! If you want to do another pull request without combining it with this one you can use a branch see [Git Branching](http://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell).
I think @AliasIO will merge it in a few days.

AliasIO:
Cheers. :)

PR #808, opened 2015-02-27
add Symfony detection by search in html part of path generated by assets...

NovikovViktor:
... manager of Symfony

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -5207,6 +5207,7 @@
 			"cats": [
 				18
 			],
+            "html": "^.*?\/web\/bundles\/.*$",
 			"implies": "PHP",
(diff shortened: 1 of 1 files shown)


AliasIO:
I think `web/bundles` is to broad a pattern. Other platforms might use the same folder structure and will be incorrectly identified as Symfony.

NovikovViktor:
@AliasIO , I agree with you but... just look at Yii detection way, when I add into any project line `Powered by Yii framework`, Wappalyzer will be think that my project use Yii framework. You able to revoke my PR.

AliasIO:
Thanks. I'd love to add Symfony but it's hard to fine a unique and accurate pattern.

PR #809, opened 2015-02-27
Add test for icons and fix icons

DaAwesomeP:
1. Verifies that there is a file and only one with an image-like extension for each app
2. Verifies that all app icons have a `*.png` extension
3. Verifies that the icons are actually PNGs (they could be the wrong file type with a PNG extension)
4. Verifies that there are no extra icons (all files are listed in `apps.json` or is `default.png`)

I have tested this is the Vagrant/Docker build, as well as the individual tests. It properly finds everything above.


AliasIO:
Nice one, thanks. I'm going to refactor the JavaScript, some things I don't think we need to test for and the two functions are illegible. I like the idea of checking the icons.

If we're putting JavaScript in the bin folder it should have a shebang (`#!/usr/bin/env node`) so it can be called directly.

DaAwesomeP:
While it may seem overkill, I am testing that the files are actually PNGs even though they have the extension because simply renaming an ICO or JPG to a PNG technically works. It will show in most browsers, but it's not at all correct

AliasIO:
See #d6debd16bbfaff3fcbd51efdb33195fd750c9505. I kept the mime type check but removed checking for extra files in the directory. You can now run "wappalyzer validate-icons" as a command. I moved the `node_modules` folder away, the `bin` folder should only contain commands.

DaAwesomeP:
The extra files check is to make sure that extra and useless bulk of the extension would not build. I had to remove 16 icons in order for my commit to pass. I'm not sure how many mb that was, but it was probably worth it. If any apps are to be removed from `apps.json` in the future, it would ensure that a proper cleanup is completed.

AliasIO:
Yeah, you're not wrong. My reasoning is that extra files don't break anything whereas missing files do. I also like to keep the shell scripts lean.

src/apps.json
```diff
@@ -5638,7 +5638,7 @@
 				23
 			],
 			"headers": {
-				"Via": "Varnish",
+				"Via": ".*Varnish",
```
gadcam: Why did you changed it ?

DaAwesomeP: I noticed that it wasn't triggering. Varnish now correctly reads `1.1 varnish` or `1.1 varnish, 1.1 varnish` on pages. The `Via` header I believe is supposed to have the HTTP version in it anyways. I notcied this multiple times when observing the headers of different websites.

AliasIO: Aren't the patterns effectively identical?

DaAwesomeP: One will find only `Varnish` and the other will find that as well as `1.1 Varnish`. It wasn't working before. I did test it.

AliasIO: In any case they should be identical so maybe it's a bug. `Varnish` is not the same as `^Varnish`. All expressions are treated the same way though so I'm not sure why this would only occur in headers.

PR #812, opened 2015-03-05
Added INFOnline

q--:
According to [their website](//www.infonline.de), they offer "Internet [Audience Measurement](https://en.wikipedia.org/wiki/Audience_measurement)". It looks like a form of analytics to me, so I put it in that category.


DaAwesomeP:
Please add another commit with `INFOnline.png` in `src/icons` that is 32x32 and losslessly compressed.

AliasIO:
Thanks!

PR #813, opened 2015-03-06
fix Enyo (js framework)

bopoda:
wrong detect here http://lenyo.com.ua/weddings/vasil-ta-karina
by `<script src="http://lenyo.com.ua/wp-content/themes/lenyo/js/lenyo.js"></script>`
@AliasIO review plz


AliasIO:
Thanks!

PR #817, opened 2015-03-11
Adding Mean.io

liorkesos:
Adding logo + mean regex (currently matching Mean 0.4.3 that contain the X-Powered-by ...


DaAwesomeP:
Could you add that MEAN implies, `AngularJS`, `MongoDB`, `Expres`s, and `node.js`?

liorkesos:
I updated the additional implies...

src/apps.json
```diff
@@ -3378,6 +3378,21 @@
...
+				12
+			],
+    			"implies": [
+    			      "MongoDB",	
+			      "Express",
+			      "AngularJS",
+			      "node.js"
+			],
+			"headers": {
+				"X-Powered-CMS": "Mean.io"
```
gadcam: You need to escape the dot like this <code>"X-Powered-CMS": "Mean\\\\.io"</code> ortherwise it will match any character.

liorkesos:
Added the escape on the "."

AliasIO:
Thanks!

PR #818, opened 2015-03-15
Mean

liorkesos:


DaAwesomeP:
You need to not have a space after `[^>]` or it will require (not be optional) to only have a space there. You also need to single escape the `src`'s double quotation mark. I would do it like the following so that it can detect both marks:
```json
"html": "<script[^>]src=('|\")\\/modules\\/aggregated\\.js",
```
Please add another commit.

DaAwesomeP:
Great! Now it passes the tests.

src/apps.json
```diff
@@ -3382,8 +3382,9 @@
 			"cats": [
 				12
 			],
+			"html": "<script[^>]src=('|\")\\/modules\\/aggregated\\.js",
```
gadcam: You should write <code>[^>]+</code> to match several characters.
You don't have to escape slashes.
The group has to be non-capturing like this <code>(?:'|\")</code> but when there is only one character I prefer to write <code>['\"]<\code> as it is easier to read.
Maybe the script is a better option.. see wiki\Specification
Then it will perfect. :) Thank you for contributing !

PR #822, opened 2015-03-19
Add Catberry.js Framework

pragmadash:
[Catberry](http://catberry.org) is a framework with Flux architecture, isomorphic web-components and progressive rendering.


AliasIO:
Thanks!

PR #825, opened 2015-03-22
Add support for hapi.js

briandela:
Adding support for trying to detect hapi.js (http://hapijs.com/).


gadcam:
Could you provide some example websites using hapi.js ? I think there are some more way to detect it. :)

briandela:
@gadcam Here are some:

- http://hapijs.com/
- https://www.npmjs.com/
- https://www.barn2door.com
- http://hood.ie/

I couldn't see anything else to identify the server definitively - it doesn't set a `server` or an `x-powered-by` header so I based it on the most commonly used CSRF and auth plugins of hapi (also built by the hapi team).

DaAwesomeP:
Some of those sites are going through proxies (Apache or Nginx), so that would remove/replace the server header. Could you run hapi locally and see what it does (i.e. follow their example where you can access it from localhost)?

briandela:
It doesn't set any server headers. I run it locally and in production day-to-day.

gadcam:
@briandela I took a look a https://github.com/hapijs/hapi to check the source code but I didn't find any obvious pattern. I think you're right : the only way to detect hapi.js is to search for popular plugins.. so good job !

briandela:
@gadcam Yep - I did the same thing re looking through the code.

briandela:
This ok to merge as is or should I look at making additional changes?

AliasIO:
Good as-is, thanks. :)

briandela:
@AliasIO Out of curiosity - just wondering when this would get released - I noticed 2.38 was release but didn't see hapi.js detection in the apps.json shipped with it.

AliasIO:
It's supposed to be in there, the Firefox version definitely has it. I'll find out what happened and push another version.

briandela:
@AliasIO Yep - saw it on firefox. Thanks for offering to push again for chrome. Much appreciated.

PR #834, opened 2015-04-03
Update Kentico CMS recognition and icon

ondrejsevcik:
Hi, I've update Kentico CMS recognition. The headers and meta recognition was removed, because html recognition works across all versions of Kentico CMS.  

Thank you.


AliasIO:
There's no reason to remove it though, especially since the generator meta tag includes version information. I'll fix the URL up though.

PR #836, opened 2015-04-07
Added CA Single Sign-On signature

atcazzual:
- Added CA Single Sign-On signature and new 'Identity Management' category for it
- Added the LiSESSIONID cookie back to the Lithium signature as it was mistakenly removed


AliasIO:
Thanks!

It may be better to split the HTML patterns up into an array.

Some more work needs to be done to add a category. All locales need to be updated, even if it's not translated. The category also needs to be assigned a priority in the Chrome and Firefox drivers, perhaps we can add it between "Marketing Automation" and "CDN".

https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/chrome/js/driver.js#L230
https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/firefox/lib/driver.js#L438

PR #838, opened 2015-04-09
Fix phantomjs problem with non ssl3 sites

m0sth8:
By default phantoms JS use SSLv3 only:
`  --ssl-protocol=<val>                 Sets the SSL protocol (supported protocols: 'SSLv3' (default), 'SSLv2', 'TLSv1', 'any')`
For this reason it doesn't work with non ssl3 sites like https://mail.ru.


AliasIO:
Excellent, thank you.

PR #839, opened 2015-04-09
Don't print status message in non verbose mode

m0sth8:


AliasIO:
That line can be deleted actually. I must have left it in there by accident.

PR #840, opened 2015-04-09
Add --resource-timeout argument

m0sth8:
Sometimes 9000 ms isn't enough


AliasIO:
I was planning on adding this, thanks!

PR #860, opened 2015-05-05
add cdnjs

PeterDaveHello:


q--:
According to @ElbertF in pull request #132, this category isn't meant for CDNs that only serve components and not the full web page.

PeterDaveHello:
hmmmm ... what about something like `web lib cdn`?

PR #862, opened 2015-05-11
Added SweetAlert icon

q--:
Created based on the logo.


q--:
@AliasIO Any idea why the tests failed? The problem does not appear to be with my commit.

AliasIO:
It looks like something broke at TravisCI's end. :)

PR #866, opened 2015-05-14
Add Orchard CMS support

urbanit:
Add Orchard CMS support


AliasIO:
Thanks!

src/apps.json
```diff
@@ -4131,6 +4131,16 @@
 			},
 			"website": "oracle.com"
 		},
+		"Orchard CMS": {
++			"cats": [
```
gadcam: "+" I think there is a little problem..

AliasIO: Not sure what that means, are they spaces?

q--: @AliasIO https://github.com/AliasIO/Wappalyzer/blob/master/src/apps.json#L4135

urbanit:
Is everything ok?

AliasIO:
Oh, right. Looks like this was copy/pasted from a diff? I'll fix it.

PR #871, opened 2015-05-24
Added Discuz! X

JLChnToZ:
[Discuz!](http://discuz.com) ([wiki](https://en.wikipedia.org/wiki/Discuz!)) is a popular forum software which is in use in many Chinese communities.


AliasIO:
Thanks!

PR #873, opened 2015-05-25
Add VTEX Integrated Store support

udlei:
Add VTEX Integrated Store Support


AliasIO:
Thanks! Can you please provide a 32x32 icon?

PR #878, opened 2015-06-02
Add Odoo

dfrankland:
Odoo is a suite of web based open source business apps. http://odoo.com


src/apps.json
```diff
@@ -3881,6 +3881,23 @@
...
+			"implies": [
+				"Python",
+				"PostgreSQL",
+				"node.js",
+				"Less"
+			],
+			"meta": {
+				"generator": "Odoo"
+			},
+			"html": "(<link[^>]* href=[^>]/web/css/(web\\.assets_common/|website\\.assets_frontend/)|<script[^>]* src=[^>]/web/js/(web\\.assets_common/|website\\.assets_frontend/))\\;confidence:50",
```
gadcam: You can use a script field for the second part of the pattern see https://github.com/AliasIO/Wappalyzer/wiki/Specification.
I think <code>[^>]</code> should be <code>[^>]+</code>

dfrankland:
@gadcam thanks for the heads up! I thought one could only match actual scripts instead of paths to scripts in the scripts field.

gadcam:
Thank you for contributing !
There is a bracket in more at the end of the script pattern, else everything is perfect.

dfrankland:
I was wondering why every page was showing 25% Confidence in Odoo lol. Removed the extra parentheses now.

AliasIO:
Thank you.

PR #879, opened 2015-06-02
Fix Travis

DaAwesomeP:
An odd character had been added to the first line and first character of `.travis.yml`, so Travis acted as if the file didn't exist.

Also, make sure that in the repo settings in Travis you set concourrent jobs to at least two and press enter (it may complain about validation, but it should work) if it still doesn't work.


AliasIO:
Thanks, I couldn't work it out. Perhaps it was a BOM character?

DaAwesomeP:
Yay! Wappalyzer is passing again.

PR #882, opened 2015-06-08
HHVM support

Ambroos:
HHVM by default adds headers like these, just like PHP:

`X-Powered-By:HHVM/3.7.1`

Using HHVM usually means you're using PHP, unless you're writing Hack (which is very unlikely), hence the 'implies'.


AliasIO:
Great, thank you!

PR #883, opened 2015-06-08
Check for `X-Powered-By: Craft CMS` header

brandonkelly:
As of Craft 2.4, Craft will send this header by default.


AliasIO:
Thanks!

PR #885, opened 2015-06-11
Added Roadiz CMS recognition

ambroisemaupate:


src/apps.json
```diff
@@ -4763,6 +4763,17 @@
...
 			"website": "ritecms.com"
 		},
+		"Roadiz CMS": {
+			"cats": [
+				1,
+				11
+			],
+			"implies": "PHP",
+			"meta": {
+				"generator": "Roadiz ([a-z0-9\\s\\.]+) - "
```
AliasIO: Can the capture group be used to identify the version number? If so, please add `\\;version:\\1` to the pattern.

ambroisemaupate: Yes, I’ll update my pull request shortly. Thanks for your suggestion

AliasIO:
Thanks!

PR #889, opened 2015-06-21
Add webpack detection

Ambroos:
When using code splitting with webpack, by default it adds `webpackJsonp` to the global scope. Used more and more nowadays, Instagram based their web platform on it, for example.


AliasIO:
Thank you!

PR #890, opened 2015-06-22
Add shopware 5 support

hlohaus:


hlohaus:
Testshop: www.shopwaredemo.de

src/apps.json
```diff
@@ -5090,7 +5090,7 @@
 			"meta": {
 				"application-name": "Shopware"
 			},
-			"script": "shopware\\.js",
+			"script": "(?:shopware|\/web\/cache\/[0-9]{10}_.+)\\.js",
```
gadcam: You don't need to escape the slashes.
If the pattern you added only matches version 5 you can add a version field see https://github.com/AliasIO/Wappalyzer/wiki/Specification#patterns

hlohaus: Ok, you mean like this?
  "script": "(?:(shopware)|/web/cache/[0-9]{10}_.+)\\.js;version:\\1?4:5",

gadcam: Yes, if you change it to
```js
"script": "(?:(shopware)|/web/cache/[0-9]{10}_.+)\\.js\\;version:\\1?4:5",
```
it would be perfect.
(To be accurate you need to escape backslashes but not slashes)

hlohaus: done :)

gadcam: Thank you for contributing !

src/apps.json
```diff
@@ -5086,11 +5086,11 @@
 			"cats": [
 				6
 			],
-			"implies": "PHP",
+			"implies": [ "PHP", "MySQL", "jQuery" ],
 			"meta": {
 				"application-name": "Shopware"
 			},
-			"script": "shopware\\.js",
+			"script": "(?:(shopware)|/web/cache/[0-9]{10}_.+)\\.js\\;version:\\1?4:5",
```
AliasIO: You're referencing a capture group which doesn't exist. Did you mean to wrap the digits in brackets? Also, you're defaulting to version 5 when no version information is present. This means you'll probably incorrectly identify Shopware 5 in certain scenarios.

hlohaus: The "/web/cache/" is new in shopware 5. the shopware.js is outdated. I think it should work.

AliasIO: Right, it makes sense now. I thought you meant to make the digits optional which would result in `/web/cache/` being the entire pattern.

AliasIO:
Thanks!

PR #894, opened 2015-06-23
Changes to LMS category expressions

ywarnier:
This change has to be verified. I have no clear idea of how to check that these will accurately match the corresponding pages.
In particular, I have modified the detector for Moodle sites, considering most of its traffic is registered as coming from http://support.godaddy.com/ and http://who.godaddy.com/ which are not even Moodle sites. This was due to a check on cookies headers (that might be validly maintained on subdomains of the same domain) which I have now removed. These should thus be spread between multiple subdomains now (hopefully), although it is possible the expression was just bogus and was generating visits for people simply going to godaddy... not sure
New detector expressions are less of an issue as they are new, but please review before merging to avoid one expression to consume a lot of resources or to check something undue.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -3738,10 +3738,7 @@
 				21
 			],
 			"env": "^moodle",
-			"headers": {
-				"Set-Cookie": "MoodleSession"
-			},
-			"html": "<img[^>]+moodlelogo",
+			"html": "^M.str\\s=\\s.*\"moodle\"",
 			"implies": "PHP",
 			"website": "moodle.org"
 		},
--- a/src/apps.json
+++ b/src/apps.json
@@ -3765,8 +3765,6 @@
 			],
 			"env": "^moodle$",
 			"html": "M\\.str = [^(moodle)]*\"moodle\"",
-			"html": "(<script|link)[^>]*mg-(core|plugins|templates)",
-			"html": "<link[^>]* href=[^>]+nv\\.d3(?:\\.min)?\\.css",
 			"implies": "PHP",
 			"website": "moodle.org"
 		},
--- a/src/apps.json
+++ b/src/apps.json
@@ -642,7 +642,7 @@
 				21
 			],
 			"meta": {
-				"copyright": ".*Blackboard Inc$"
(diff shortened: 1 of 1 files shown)


AliasIO:
It didn't pass the automatic validation, have a look at the Travis CI report. It returned the following:

`Avoid ".+" and ".*" in HTML patterns. Consider using "[^>]+" or "[^<]+" instead.`

This is to keep pattern matching within a tag and avoid matching plain text.

ywarnier:
How would you suggest the regexp for this:
```
<div id="registration_footer">
    <div class="registration-content">
        <div class="left">&copy; 2015 <a href="http://www.instructure.com">Instructure</a></div>
```
Clearly, matching only the <a> tag would be too wide (many sites can have a link to this), but including the 3 divs would require matching a line ending + several < and > + a &copy; sign which I'm not sure how to escape... :-)

AliasIO:
The copyright notice is good enough, something like this will do:

`&copy; 20\\d{2} <a href="http://www\\.instructure\\.com">Instructure</a>`

You can match text outside tags. The reason `.+` should be avoided is because it could match the entire page. I also prefer avoiding multi-line patterns for performance reasons.

ywarnier:
I have another issue with Moodle... I need to recognize something like this:
```
//<![CDATA[
M.str = {"moodle": ... };
//]]>
</script>
```
I'm trying with "html": "M\\.str = [^(moodle)]*\"moodle\"", but I get an error that html tags require a < or >... Without including a multiple-line pattern, I cannot do that. Any suggestion?

AliasIO:
Normally you'd be able to use the JavaScript variable (i.e. `"env": "^M$"`), however `M` is too short to be reliable. The snippet you posted isn't really usable with the current implementation.

ywarnier:
That's pretty much the *only* common element on all Moodle portals. The current expression used is incorrect (matches sites that have nothing to do with it).

AliasIO:
Alright, let's try a newline. Something like `<!\\[CDATA\\[\\n\\s*M\\.str = {\"moodle\"`. If you can, please make sure this pattern works without any other patterns.

PR #898, opened 2015-06-26
Fix regex compatibility issues with PCRE

fxfactorial:
These regexes are not compatible with PCRE based libraries


AliasIO:
Thanks for the corrections!

PR #901, opened 2015-06-30
add google web server

redabhr:


klonos:
...you might wanna change ```fr.wikipedia.org/wiki/Google_Web_Server``` to ```en.wikipedia.org/wiki/Google_Web_Server``` (English instead of French) or plain ```wikipedia.org/wiki/Google_Web_Server```.

redabhr:
yep, right

AliasIO:
Thanks!

PR #904, opened 2015-07-02
Add Landing Page Builder Category / Put Unbounce In It

mattcoady:


AliasIO:
To add a category the locales need to be updated, as well as references to categories in the drivers. We can probably add this one between 'Message Board' and 'Wiki' (order of importance).

https://github.com/AliasIO/Wappalyzer/tree/master/src/drivers/chrome/_locales
https://github.com/AliasIO/Wappalyzer/tree/master/src/drivers/firefox/locale
https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/chrome/js/driver.js#L230
https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/firefox/lib/driver.js#L438

PR #909, opened 2015-07-11
Implement Yoast SEO recognition

jdevalk:
[Yoast SEO](https://yoast.com/wordpress/plugins/seo/) is a WordPress plugin that's used on approximately 4 million websites. 

I didn't add the "implies" tag on purpose as we're currently in the process of porting Yoast SEO to other platforms too.

I chose marketing automation as a category as that seems the best fit but am open to suggestions :)


AliasIO:
Thanks Joost! I was actually reading your article on Magento SEO the other day. Greeting from a fellow Dutchie.

PR #912, opened 2015-07-11
Added http-kit

kolen:
[HTTP Kit](http://www.http-kit.org/) web server.

It has no logo or favicon, so there's no icon yet (I'm not sure if it's ok to have detection rule without icon).


AliasIO:
The build failed, can please add a 32x32 PNG icon?

kolen:
Added generic cube icon, because http-kit has no logo.

AliasIO:
Thanks!

PR #913, opened 2015-07-12
Changes for Yesod/Warp

### LINKED ISSUE #910: "Warp" false-positive detections because matches substring
Detects [here](http://webmail.vasslatam.com/webmail/), but there's IceWarp ``Server:IceWarp/11.1.2.0``, different product.

```
		"Yesod": {
			"cats": [
				18,
				22
			],
			"headers": {
				"Server": "Warp/\\d+(\\.\\d+)+"
			},
			"implies": "Haskell",
			"website": "www.yesodweb.com"
		},
```

Maybe regex for headers should match whole string, not substring, so it will be not required to add ``^`` and ``$`` in every regex in ``apps.json``?

Also, Warp does not mean Yesod, but it's another issue.

kolen:
- Changed name to Warp, because this detection rule is for Warp, not Yesod
  
  Warp is generic web server and it does not imply Yesod. It was a part of Yesod project as simple development server, but now it's production-ready webserver, and it's generic, not tied to Yesod framework.

  I have no idea how to detect Yesod, so I didn't create detection rules for it.

  Left the same icon: Warp was created by Yesod team and it has no logo, so I think Yesod logo fits better than generic "box" icon.

- Removed category 18 "web-frameworks"
- Prevent matching with prefix (added ``^``) - because of false detections of IceWarp webmail (fixes #910)
- Added version matching


AliasIO:
Thanks!

PR #917, opened 2015-07-17
NPM driver detect from html

Johanndutoit:
First off, awesome lib ! Wanted to call this directly from one of our NodeJS workers but we already have the request data so wanted a way to just pass that data in. Then saw the #detectFromHTML method that was still stubbed, so added in logic for the method and kept the names.

Also tried to clean up a little bit and make it all happen through one function.

Checked tests and added in a test to check the new method. Using HTML and headers I got from the site that is currently configured to be used for testing - http://codelanka.github.io/Presentation-Engines

Let me know if I missed something.

### DIFF
--- a/src/drivers/npm/index.js
+++ b/src/drivers/npm/index.js
@@ -4,26 +4,65 @@ var request = require('request');
 var fs = require('fs');
 var path = require('path');
 
-//TODO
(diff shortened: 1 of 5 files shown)


AliasIO:
Nice one, thank you!

PR #918, opened 2015-07-18
Updated to use the correct environment variable

Johanndutoit:
Saw that I forgot to actually check for the environment variable. "testing" is only set when testing and makes the tests check 2 folders up rather than in the same folder for the module as it would be when packaged on npm.

What is the process on pushing new versions to NPM ? Would love to start using it and 2 packages for the same project would just confuse people.

### DIFF
--- a/src/drivers/npm/index.js
+++ b/src/drivers/npm/index.js
@@ -94,7 +94,7 @@ function runWrappalyer(options, data, cb) {
 	var wappalyzer = null;
 
 	// change depending on the environment
-	if( process.env.NODE_ENV ) {
+	if( process.env.NODE_ENV == 'testing' ) {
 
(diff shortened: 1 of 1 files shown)


AliasIO:
There is no official NPM package as of yet. I just signed up an account on npmjs.com but I haven't used it before.

Johanndutoit:
@AliasIO do me a favour and add me on google.com/+Johanndutoit ( assuming you have GMAIL here :) ) then we can chat and get this baby on NPM. Going to commit a few lines to the build script to add npm on that list too. Then you should just be able to "npm publish" and we/everyone can start using.

AliasIO:
Please email me at info at wappalyzer.com or elbert at alias.io. I'm not on Google +, Facebook et al.

Johanndutoit:
Mailed :+1:

PR #928, opened 2015-08-06
Added Wikispaces

kolen:
http://www.wikispaces.com/ hosted service. Can appear on custom domains (example: http://clojure.org).


AliasIO:
The tests failed on this one; all HTML patterns should have at least an opening or closing tag in them to prevent them from matching plain text. Instead of `class` you could have `<[^>]+class`, for instance.

kolen:
Added matching for start tags.

AliasIO:
Thanks!

PR #929, opened 2015-08-08
Update Grav generator

dfrankland:
Changed with this commit: https://github.com/getgrav/grav/commit/b2cd4db3955871715ca759f288a885ed44d57b2e

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -2295,7 +2295,7 @@
 				"PHP"
 			],
 			"meta": {
-				"generator": "Grav(?:\\s([\\d.]+))?\\;version:\\1"
+				"generator": "GravCMS(?:\\s([\\d.]+))?\\;version:\\1"
 			},
 			"website": "getgrav.org"
 		},


src/apps.json
```diff
@@ -2295,7 +2295,7 @@
 				"PHP"
 			],
 			"meta": {
-				"generator": "Grav(?:\\s([\\d.]+))?\\;version:\\1"
+				"generator": "GravCMS(?:\\s([\\d.]+))?\\;version:\\1"
```
AliasIO: Should we make `Grav(?:CMS)?` in order to match existing installations?

dfrankland: My only reason for not adding `Grav(?:CMS)?` is that Grav hasn't even reached 1.0 and also https://wappalyzer.com/applications/grav shows many websites not actually using Grav the CMS for some reason... However, besides that, I think adding `Grav(?:CMS)?` wouldn't be bad :) What are your thoughts?

AliasIO: Let's not make it optional in that case, the more specific the better.

AliasIO:
Thanks!

dfrankland:
Cheers! :+1:

PR #935, opened 2015-08-18
adding osticket.com

berezuev:
I've added a popular helpdesk crm "osticket"
Their github repo is: https://github.com/osTicket/osTicket-1.8


AliasIO:
Would this fit in 'Issue Trackers'? There is a bit more work involved in adding a new category.

berezuev:
ok, you're right. I've just missed this category

PR #939, opened 2015-08-22
Added DM Polopoly

q--:
[Their website](www.atex.com/products/dm-polopoly) lists some other technologies supposedly used by the CMS, such as Tomcat, but I'm not 100% sure the text means that they are used at all times. Let me know if you think I should add them to the `implies`.

I'm checking *only* for paths *starting with* `/polopoly_fs/` so that websites embedding resources from websites with the CMS won't get matched. Of course, there is a risk that some Polopoly users go undetected because they're using the full URL everywhere, but I haven't seen this in practice.


AliasIO:
Thanks mate!

PR #960, opened 2015-09-21
Check if patterns are valid regular expressions by creating a new RegExp object for each of them

q--:
Surprised you weren't doing this already. I will also do a pull request for the error I've found.


AliasIO:
Me too, thanks!

PR #968, opened 2015-09-26
Validate capture groups

q--:
Detect when a version string has too little, too much or the wrong capture
groups.

Upgrade Node.JS to 4.10 because a recent version is required for the new
code.

I probably spent _way_ too much time on this&hellip;

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "0.10"
+  - "4.1.0"
 sudo: required
 before_install:
   - export WAPPALYZER_ROOT=$TRAVIS_BUILD_DIR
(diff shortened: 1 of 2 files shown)


DaAwesomeP:
The changes to `.travis.yml` now conflict with #969.

Not only have other parts of the file changed (git sometimes does and doesn't merge this type of conflict correctly), but I have switch it to NodeJS 0.12.x. Your commit switches the Travis one to io.js and does not change the Docker build. If you do upgrade the Docker build, then you will need to change the package that it is installing as well as the APT repo from here: https://nodesource.com/blog/nodejs-v012-iojs-and-the-nodesource-linux-repositories

AliasIO:
Merged offline in 61d5ffbcfe1b23625cb46be4fb9310047c4f41b9. I have commented out the validate-regex script for now until apps.json is actually passing the tests.

PR #969, opened 2015-09-26
Compress icons with build, use new Travis container, update NodeJS version

DaAwesomeP:
- Compress images with every build to ensure minimum file sizes
 - Update to NodeJS 0.12 (the NodeJS version available in the standard apt repo is now insanely old)
 - Remove use of  `sudo` in Travis and set `sudo: false` for the [new container](http://docs.travis-ci.com/user/migrating-from-legacy/).
 - Add more exclusions to `.gitignore`


AliasIO:
Thanks!

PR #979, opened 2015-10-08
Updated website URL WP-Super-Cache

janyksteenbeek:
Updated website URL WP-Super-Cache. ocaoimh.ie/wp-super-cache/ is redirecting to new URL z9.io/wp-super-cache/, changed it just in case they remove their old domain.


AliasIO:
Thanks!

PR #998, opened 2015-10-18
Removed Google Analytics from Chrome

rugk:
Closes https://github.com/AliasIO/Wappalyzer/issues/994

I hope it works this way... :smiley:


AliasIO:
Thanks!

PR #1019, opened 2015-11-27
Updated Thelia specification

bibich:
Updated Thelia specification

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -6541,17 +6541,12 @@
         1,
         6
       ],
-      "headers": {
-        "Set-Cookie": "thelia_cart="
-      },
+      "html": "<(?:link|style|script)[^>]+/assets/frontOffice/",
       "implies": [
         "PHP",
         "Symfony"
       ],
-      "meta": {
-        "generator": "Thelia v([\\d.]+)\\;version:\\1"
-      },
-      "website": "www.thelia.net",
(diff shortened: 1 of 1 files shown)


q--:
Are you sure the Theila version matching the patterns you removed is in use _nowhere_?

bibich:
In fact, the version can't be detected, so the removed pattern was useless.

bibich:
Is there something wrong with this PR, an update to do ?

AliasIO:
No, I've just been slack with merging. Thank you. :)

bibich:
No problems :)
Thanks.

gadcam:
It's a good thing that you updated the patterns but I don't understand why did you remove the former patterns ? (They both match on http://demo.thelia.net/.)

bibich:
Yes it matches on http://demo.thelia.net/ because this site uses our default template (theme).
But in most cases, Thelia websites don't use this template or remove the mata tag generator.
 
And for the second pattern (the Set-Cookie), it could be deactivated, or it happens only on rare occasions.

gadcam:
Ok , I didn't know that but it's same for any pattern: if someone wants to hide all the tools powering a website from Wappalyzer he can.
In my opinion the better would be to keep all the patterns: it costs us almost nothing and it would improve the detection rate.

PR #1020, opened 2015-11-27
Add MaxCDN

Ambroos:
The enterprise side of MaxCDN used to be named NetDNA, and it’s still what they use in their headers. Look at www.canvas.be for an example of a full website going through MaxCDN.


AliasIO:
Thanks!

PR #1023, opened 2015-11-30
Enable capture group validation, fix unnecessary capture groups and add missing a missing capture group for Oracle Commerce Cloud

q--:
I also removed the lowercase-named `penguin.png` file, as it was causing problems. The uppercase version is still there.


AliasIO:
Thanks!

PR #1035, opened 2015-12-21
Add Bounce Exchange

q--:


src/apps.json
```diff
@@ -822,6 +822,16 @@
       "website": "cibonfire.com",
       "icon": "Bonfire.png"
     },
+    "Bounce Exchange": {
+      "cats": [
+        32
+      ],
+      "env": "^bouncex$",
+      "html": "<script[^>]*>[^>]+\\.src\\s*=\\s*['\"](?:https?:)?//tag\\.bounceexchange\\.com/",
```
gadcam: Why do you write this pattern ? Isn't it the same than the script's one ?

q--: BounceX is loaded by this script:

	
	<script>
	(function() {
		function async_load(){
			var s = document.createElement('script'); s.type = 'text/javascript';
			s.src = '//tag.bounceexchange.com/1194/i.js';
			var c = document.getElementsByTagName('script')[0]; c.parentNode.insertBefore(s, c);
		}
		if(window.attachEvent) { window.attachEvent('onload', async_load); }
		else { window.addEventListener('load', async_load, false); }
	})();
	</script>
	

As you can see, the *script* tag that loads the actual BounceX code from *tag.bounceexchange.com* is added to the document with JavaScript. Without this code, BounceX wouldn't be detected in environments that don't execute JavaScript.

The part of the script I'm trying to match is this:

    .src = '//tag.bounceexchange.com/

gadcam:
You're right. I read too fast, my mistake.

src/apps.json
gadcam: I read to fast and didn't know it could be loaded like this.
:+1:

q--: No problem.

P.S, merry Christmas and a happy New Year, everyone.
Op 25 dec. 2015 19:16 schreef "Camille Barneaud" <notifications@github.com>:

AliasIO:
Thanks!

PR #1039, opened 2015-12-24
Add Sails.js

DaAwesomeP:
I categorized Koa and Express as web servers as well.


DaAwesomeP:
The build errored. That's odd. Seems to have been unable to download something.

src/apps.json
```diff
@@ -5663,6 +5665,21 @@
       "website": "www.sbuilder.ru",
       "icon": "S.Builder.png"
     },
+    "Sails.js": {
+      "cats": [
+        18
+      ],
+      "headers": {
+        "Set-Cookie": "^sails.sid$",
```
gadcam: If you want to match a dot I think you'd better escape it.

DaAwesomeP: My bad. I'll fix it.

src/apps.json
```diff
@@ -5663,6 +5665,21 @@
       "website": "www.sbuilder.ru",
       "icon": "S.Builder.png"
     },
+    "Sails.js": {
+      "cats": [
+        18
+      ],
+      "headers": {
+        "Set-Cookie": "^sails\\.sid$",
```
DaAwesomeP: @gadcam Fixed

AliasIO:
Thanks!

PR #1041, opened 2015-12-27
Add support for svg icons in drivers

pallxk:
This adds support for svg icons in following drivers:
* Mozilla Firefox
* Google Chrome
* Bookmarklet
* HTML

Works in address bar and popups.

### DIFF
--- a/bin/wappalyzer-links
+++ b/bin/wappalyzer-links
@@ -30,19 +30,23 @@ echo "Creating hard links..."
 ln -f $path/wappalyzer.js $path/drivers/firefox/lib
 ln -f $path/apps.json     $path/drivers/firefox/data
 ln -f $path/icons/*.png   $path/drivers/firefox/data/images/icons
+ln -f $path/icons/*.svg   $path/drivers/firefox/data/images/icons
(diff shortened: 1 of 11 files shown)


DaAwesomeP:
@pallxk Great! I see that you added support for the new `icon` attribute.

@AliasIO Could you please drop the SVG for the Wappalyzer logo into the icons folder as `default.svg`? After #1039 I'm going to change all of the no-icon ones with the Wappalyzer icon to that as well as a lot of the popular apps (jQuery, Nginx, Angular, Bootstrap, etc.).

AliasIO:
Thanks!

@DaAwesomeP I don't have the Wappalyzer icon in vector but do have it in a reasonably high res.

DaAwesomeP:
@AliasIO OK, if you could please drop that as `default.png`, that would be great! The [current one](https://github.com/AliasIO/Wappalyzer/blob/master/src/icons/default.png) is only 32x32.

AliasIO:
There are more sizes here: https://github.com/AliasIO/Wappalyzer/tree/master/src/drivers/firefox/images

PR #1046, opened 2016-01-05
Lower confidence for all basket.js patterns

q--:
The patterns for basket.js are quite generic and match at a lot of sites that aren't using this script. The confidence should be lower.


AliasIO:
Thanks.

PR #1047, opened 2016-01-05
Re-add version detection for Arastta

q--:
Version information in the headers was removed in version 1.2.1, but not
all websites are using this version. Additionally, an empty "X-Arastta"
header implies Arastta 1.2.1+ so detection was added for this as well.


AliasIO:
Thank you.

PR #1050, opened 2016-01-07
Respect "Show upgrade message" option in Chrome

bfred-it:
Local Storage saves everything as strings: `"0"` is not falsey, `0` is.

Until now Chrome always opened a tab when Wappalizer updated, regardless of that option.


AliasIO:
Odd, `"0" == false` returns `true` but `"0" ? true : false` returns `true` as well.

PR #1057, opened 2016-01-11
Create anchorCache.png

ioli:

PR #1064, opened 2016-01-12
Remove many icon duplicates, change some icons to SVG

DaAwesomeP:
I'll do another pull like this again. This one:
 - Removes many large and obvious icon duplicates
 - Makes use of only having one default icon file
 - Changes some icons to SVGs


gadcam:
Apache.svg: I would say why not a full and beautiful logo however how would you display it in the URL bar ?

DaAwesomeP:
@gadcam It's the same icon that was the PNG. It's not square, and I haven't tested how it scales (we might need to only set the `height` attribute on the `` for proper scaling, but it is locked into a certain size (see [`drivers.js`](https://github.com/AliasIO/Wappalyzer/blob/71a0f3effc30f29fbc9a7e1d7e412d82a08909e2/src/drivers/html/js/driver.js#L54)). The SVGs are fantastic for higher-resolution displays (i.e. retina).

AliasIO:
Nice one, thanks!

DaAwesomeP:
I've reported the problem with the tests here: https://github.com/sindresorhus/file-type/issues/48. When a fix is released, I'll update the version.

DaAwesomeP:
@AliasIO The culprit: anchorCache.png. I'll submit a fix to catch this error.

@ioli Please do not upload blank files. I'm guessing by the name of the pull (#1057) that you did it through the GitHub web interface. To properly upload non-textual file (binary, like images), you must use the `git` CLI. On the plus side, you helped strengthen the tests to prevent this.

DaAwesomeP:
See #1065.

ioli:
sorry please fix problem error , sorry  thanks

gadcam:
@DaAwesomeP Yes that was the question I was raising : How do we manage the scaling to keep it "clean" and visible ?
How about the size of apache.svg also ? (93.1 kb) Could we compress it or the Apache license is forbidding it ?
For the remainder well done, you did a great job!

ioli:
sorry for big problem , but i dont understund add image to github !!

how to add my pc to github icon ?  variable ?

DaAwesomeP:
@gadcam It should be compressed in the build process. I'm sure that yes, it's 93 kilobytes probably outweighs the 1.3 kilobytes that the several duplicate ones were. However, the [build size for both the Firefox and Chrome extensions are still each less than one megabyte](https://travis-ci.org/AliasIO/Wappalyzer/builds/102020390#L331-L335) (this is why I have `ls` run afterwards in the Travis build). I'm not sure how much we should be concerned with the size (just my opinion).

gadcam:
@DaAwesomeP Then it's perfect!
In my opinion, by keeping it light we enable Wappalyzer to grow over and over and when Wappalyzer is used as a dependancy it's also important to keep it convenient.
PS: a slow connection is not that uncommon ;)

PR #1085, opened 2016-01-28
bugfix at nginx server detection

adibalcan:


src/apps.json
```diff
@@ -4628,7 +4628,7 @@
 				22
 			],
 			"headers": {
-				"Server": "nginx(?:/([\\d.]+))?\\;version:\\1"
```
q--: Why should Wappalyzer stop detecting the NGINX version?

adibalcan: last nginx version haven't version in server header

src/apps.json
```diff
@@ -4628,7 +4628,7 @@
...
+				"Server": "nginx.*"
```
gadcam: Both patterns will match the same thing as .* is "optional".

src/apps.json
gadcam: The <code>?</code> is making it optional.

q--: @adibalcan The version is still shown for version 1.9.8 at http://nginx.org/. Also, as @gadcam pointed out, your change doesn't improve detection; [`?`](http://www.regular-expressions.info/optional.html) after the group makes it optional. Furthermore, even if a current version of the application doesn't reveal the version number, that's still no reason to kill version detection for older versions.

PR #1086, opened 2016-01-28
Update apps.json

siparker:
fixed google tag manager detection

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -2636,8 +2636,8 @@
 			"cats": [
 				42
 			],
-			"env": "^googletag$",
-			"html": "googletagmanager\\.com/ns\\.html[^>]+></iframe>",
+			
+			"script": "//www\\.googletagmanager\\.com/gtm\\.js",
 			"icon": "Google Tag Manager.png",
 			"website": "www.google.com/tagmanager"
 		},


src/apps.json
```diff
@@ -2636,8 +2636,8 @@
 			"cats": [
 				42
 			],
-			"env": "^googletag$",
-			"html": "googletagmanager\\.com/ns\\.html[^>]+></iframe>",
```
AliasIO: Did you mean to remove these?

siparker: yes the env seemed to make no difference in detection so saved the line to keep it as small as possible. 
the html line was not working at all and never detected anything.

AliasIO: If the `env` works than we should keep it in, it will identify the app when it's loaded through a different script.

siparker: i did test with just the env in and it seems to do nothing?
in fairness i dont know how to check env variables properly though so if you think it should stay in by all means please just edit the html to be the script one instead.

q--: @siparker You can see if an environment variable is present simply by typing its name into your  browser console. If it's autocompleted, then the variable is present. If it isn't, then the variable is not present but you'll still want to make sure that the variable is present *nowhere*, even in older versions and/or on different pages. Also, Wappalyzer patterns are case-insensitive but variable names _are_!

And as a rule of a thumb: don't remove detection patterns unless you're absolutely, positively sure it does not manage to detect a single version of the targeted application (even if that version was deprecated several years ago) or if you've written a replacement rule that detects the _same_ property better.

siparker: hi. thats great info. would def be worth adding to the wiki!!

in that case the env variable is indeed present on the pages i have to test. should i try to modify the pull request or did you just want to get that one line?

Also any other recommendations on how to identify possible env variables if you dont know what they are called?

PR #1090, opened 2016-02-01
Add Segment Detection

st3w4r:
Segment load their files from CDN `cdn.segment.com`
And their `Analytics.js` library is loaded.
[Segmnet Analytics.js](https://segment.com/docs/libraries/analytics.js/)
Thanks for the review.


AliasIO:
Thanks!

PR #1092, opened 2016-02-04
Added Stamplay service

mirkods:
Added Stamplay service


AliasIO:
Thanks!

mirkods:
When will be available with the new version?

PR #1105, opened 2016-02-24
Fix Google Analytics and Yandex.Metrika patterns

kudmni:


q--:
Are you sure the change to the GA pattern is even necessary?

kudmni:
Current GA pattern does not cover most popular  cases:
http://www.google-analytics.com/ga.js
https//ssl.google-analytics.com/ga.js
//www.google-analytics.com/analytics.js

q--:
Simple testing (using the code below) shows that the first two are matched. As for the last one, I’d argue it’s the responsibility of Wappalyzer to make sure the patterns are matched against the full URL. So if Wappalyzer doesn’t handle protocol-relative URLs correctly, that’s a bug in Wappalyzer and if you’re sure that’s the case, you may want to file an issue for that.

    "https://ssl.google-analytics.com/ga.js".match(new RegExp("^https?://[^\\/]+\\.google-analytics\\.com\\/(?:ga|urchin|(analytics))\\.js"))

kudmni:
Yes, most likely it's a bug in wappalyzer. I'll revert GA pattern in my pull-request.

PR #1121, opened 2016-03-19
Update IP.Board/IPS Suite detection

MartinAronsen:
IPB (now IPS Suite) version 4 doesn't use the global "IPBoard" object.


AliasIO:
Thanks!

PR #1149, opened 2016-05-13
Fix Underscore.js URL

leotm:
Fix [old](http://documentcloud.github.com/underscore) dead link to the [new](http://underscorejs.org/) one.
As seen on documentcloud.org/opensource.


AliasIO:
Thanks!

PR #1152, opened 2016-05-15
Add Hugo

revelt:
Let's add Hugo to the list


AliasIO:
Thanks!

revelt:
@AliasIO no problem! By the way, do you know, why Hugo does not appear in the CMS'es list: https://wappalyzer.com/applications — is it because there are too little stats counted, or is page broken and needs updating?

PR #1162, opened 2016-05-24
Added magento2 support

timneutkens:
Hi, I just added 2 regexes to support Magento 2.

The first one is in the html node and looks for `<script type="text/x-magento-init">` the regex seems broken so would appreciate if anyone could take a look at it . This would then be the leading check for Magento 2 ( It's unique.. what's more to say ;-) ) 

The second checks for `static/_requirejs` in script tags, which is kind of unique to Magento 2 (haven't seen it before) but I don't think it should be the leading check. So I added it at 50% confidence.

We could create a third check which looks for `static/frontend` but I find that way too generic. So we might add it at 20% or lower.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -4094,17 +4094,21 @@
 		},
 		"Magento": {
 			"cats": [
-				6
+			  6
 			],
-			"env": "^(?:Mage|VarienForm)$",
+			"env": [
+			  "^(?:Mage|VarienForm)$"
+			],
 			"headers": {
-				"Set-Cookie": "frontend=\\;confidence:50"
+			  "Set-Cookie": "frontend=\\;confidence:50"
 			},
+			"html": "<script [^>]*type=\"[^\"]+text/x-magento-init\\;version:\\2",
 			"icon": "Magento.png",
 			"implies": "PHP",
 			"script": [
-				"js/mage",
-				"skin/frontend/(?:default|(enterprise))\\;version:\\1?Enterprise:Community"
+			  "js/mage",
+			  "skin/frontend/(?:default|(enterprise))\\;version:\\1?Enterprise:Community",
+			  "static/_requirejs\\;confidence:50\\;version:\\2"
 			],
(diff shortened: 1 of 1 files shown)


timneutkens:
@AliasIO Do you have any thoughts on this?

AliasIO:
Hi,

I agree with what you said and the regex looks right. By broken, do you mean it's not displaying the version number as '2'? It could be that the other patterns override the version number.

timneutkens:
Hey, thanks for getting back at this. What I was trying to say is that [this](https://github.com/AliasIO/Wappalyzer/pull/1162/commits/59a51d870205e4525b9baca8eb905991f5bbe103#diff-557d7d60c04092a680ce152f367618c0R4105) regex doesn't seem to give results when I try to use it with the phantomjs driver. [This](https://github.com/AliasIO/Wappalyzer/pull/1162/commits/59a51d870205e4525b9baca8eb905991f5bbe103#diff-557d7d60c04092a680ce152f367618c0R4111) check for `static/_requirejs` is functioning properly.

timneutkens:
@AliasIO Hi, do you have an update on this? Thanks :)

timneutkens:
Hey @AliasIO, just took another look. Phantomjs renders all javascript (doh). The Magento 2 require.js library strips all x-magento-init tags. That's why the check doesn't work. Will have to do some other check. Looking for a better one right now.

timneutkens:
Fixed the regex. Current build is failing because yepnope.js is missing an icon.

timneutkens:
Hey @AliasIO, could you trigger a rebuild. I'm curious if it will pass this time since the last time it failed it was because of the yepnope issue you merged today. Thanks 👍

AliasIO:
Thanks!

timneutkens:
@AliasIO when do releases for chrome happen? So I'll know when this detection is available. Thanks for merging!

timneutkens:
Nevermind, already noticed it's been released 👍

PR #1165, opened 2016-06-11
Rename SEOshop to Lightspeed eCom and better detection for it

timneutkens:
SEOshop has been bought by Lightspeed and renamed it to Lightspeed eCom so I renamed it and added a svg logo.
I changed the way Lightspeed eCom is detected a little. Before it would only detect the admin panel of SEOshop. Now it will detect all sites using the software. .rain is their own template language based on twig. So I'm pretty sure it won't be used anywhere else.


AliasIO:
Thanks!

PR #1169, opened 2016-06-13
Minor fix on Hippo CMS detection

joostoudeman:
Hippo CMS rexex extra space caused to miss out on CMS detection if the space is missing before the html tag.


joostoudeman:
Is there any reason this pull request is not accepted yet?

AliasIO:
No reason. Thanks! :)

PR #1170, opened 2016-06-13
Add SVG icon for Shopify

admhlt:
Shopify's logo was updated a while ago, and the new version never made it into Wappalyzer. This PR removes the old (.png) logo and replaces it with the new logo in .svg format.


AliasIO:
Thanks!

PR #1174, opened 2016-06-13
Fix DoubleClick Icon

DaAwesomeP:
I must have forgotten to check this SVG before I committed it. The width was too wide and there was remnants of hidden text. It's fixed now. The other new SVG icons look fantastic though!


AliasIO:
Thanks!

PR #1176, opened 2016-06-14
Added Lighty PHP framework to the list

youssefhad:
Added Lighty PHP framework to the list


youssefhad:
did you accept contributing or not ? or we should just wait :)

AliasIO:
Thanks!

PR #1177, opened 2016-06-15
Remove version number from Arastta e-Commerce

burakcakirel:
It always shows "Arastta 1.2.1+", after this change, it shows only "Arastta" if Arastta version is greater than 1.2.1


AliasIO:
Thanks!

PR #1182, opened 2016-06-23
add arvancloud.com

MasoudNikoomanesh:
ArvanCloud Content Delivery Network (CDN)


src/apps.json
```diff
@@ -9038,7 +9038,18 @@
...
 			"website": "shopery.com"
-		}
+		},
+		"ArvanCloud": {
+			"cats": [
+				31
+			],
+			"env": "^ArvanCloud$",
+			"headers": {
+				"AR-PoweredBy": "Arvan Cloud (arvancloud.com)"
```
gadcam: You should escape the dot and the brackets by adding <code>\\\\</code> before them.

MasoudNikoomanesh: Thanks, I changed the header.

AliasIO:
Thanks!

PR #1205, opened 2016-07-23
Added support for Hotjar

joshthewebgeek:


src/apps.json
```diff
@@ -3044,6 +3044,15 @@
 			},
 			"website": "hotarucms.org"
 		},
+		"Hotjar": {
+			"cats": [
+				10
+			],
+			"env": "^hj$",
```
AliasIO: This is too generic unfortunately. Minified code on website often have two letter variables that would result in a false negative.

src/apps.json
```diff
@@ -3044,6 +3044,15 @@
...
+			"html": "h\,o\,t\,j\,a\,r",
```
AliasIO: This pattern should include HTML markup, something that can't match plaintext and only results in a detection on websites that use the application.

PR #1220, opened 2016-08-09
Update wappalyzer.rb

mossharris:
fix SSL_connect returned=1 errno=0 state=error: certificate verify failed (OpenSSL::SSL::SSLError)


AliasIO:
Thanks!

PR #1227, opened 2016-08-13
Add support for Refinery CMS

bricesanchez:


bricesanchez:
friendly ping :)

src/apps.json
```diff
@@ -5816,6 +5816,20 @@
...
+				1
+			],
+			"icon": "RefineryCMS.png",
+			"implies": "Ruby",
+			"html": [
+				"<(?:link|a href) [^>]+canonical",
+				"<[a-z]+ [^>]*id=\"page_container\"",
+				"<[a-z]+ [^>]*id=\"body_content\"",
+				"<[a-z]+ [^>]*id=\"side_body\""
+			],
```
AliasIO: These HTML patterns are too generic and will result in false positives (Wappalyzer tries to match any, not all). The first pattern will match any website that uses canonical URLs.

PR #1237, opened 2016-08-29
Use headers instead of whole response (npm driver)

kevNii:
Using the actual header-data allows to detect nginx, apache, etc. correctly now.


AliasIO:
Thanks!

PR #1251, opened 2016-09-24
Node.js is a runtime environment (new category)

tclausing:
Not a programming language.


AliasIO:
Thanks. You are correct but until we have more applications that fit in the 'runtime environment' category I'd like to leave it as-is. When comparing server-side languages it's useful to have Node.js in there even though it should technically just say 'JavaScript'.

PR #1255, opened 2016-09-27
add peyvand cms

peyvand:


AliasIO:
The tests are failing because it's missing an icon for the application. Can you please add one?

gadcam:
@peyvand I'm not sure I understood what do you have in mind making this pull request.
"Peyvand CMS" is not a CMS, is it ?
In my opinion you can not use Wappalyzer to advertise your business but @AliasIO is the only one to make decisions here.

ScionOfBytes:
@AliasIO If @peyvand Isn't gonna add an icon, can we consider this issue closed? It's been a while.

PR #1271, opened 2016-11-06
correction data

alfduca:
correction "script": "cdn\\.e-commerceparaguay\\.com", to "url": "cdn\\.e-commerceparaguay\\.com",

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -2033,7 +2033,7 @@
 				6
 			],
 			"icon": "eCommercePy.png",
-			"script": "cdn\\.e-commerceparaguay\\.com",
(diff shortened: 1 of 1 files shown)


AliasIO:
`url` only works when you're viewing the URL directly, probably not what you have in mind.

alfduca:
sites have to have if or if a link to the url (Example: http://cdn.e-commerceparaguay.com/xxxxxx) is like this? or am I wrong?

AliasIO:
It only matches if that URL is displayed in the URL bar.

ScionOfBytes:
@AliasIO @alfduca Can we consider this issue resolved?

PR #1277, opened 2016-11-16
add Kamva

hadifarnoud:
Kamva is an ecommerce platform like Shopify. This pull request will detect sites that are built with Kamva


hadifarnoud:
example site to test this on: http://sanjaaq.com/

ScionOfBytes:
@AliasIO I know I'm turning up on all these issues 😅  but I love the project and wanna see more apps being detected.

If this is good, merge?

src/apps.json
```diff
@@ -3665,6 +3665,15 @@
 			"script": "cf\\.kampyle\\.com/k_button\\.js",
 			"website": "www.kampyle.com"
 		},
+		"Kamva": {
+			"cats": [
+				6
+			],
+			"env": "^Kamva$",
+			"html": "<script[^<]+cdn\\.mykamva\\.ir",
```
gadcam: Why don't you use a script field ? Cf https://github.com/AliasIO/Wappalyzer/wiki/Specification

hadifarnoud:
@gadcam thank you for bringing this to my attention. I added a meta tag as well to detect all sites made by Kamva.

hadifarnoud:
any idea why tests fail @gadcam ?

ScionOfBytes:
@hadifarnoud I think it's because the `generator` property needs to have a string value.

hadifarnoud:
so the docs are out of date? I just followed what was said [here](https://github.com/AliasIO/Wappalyzer/wiki/Specification):
```
"meta":    { "generator": [ "Application Name", "Alternative Application Name" ] },
```

Isn't `"Kamva"` and "Camva" string?

ScionOfBytes:
@hadifarnoud Could be. From what I understand of the rules [here](https://github.com/AliasIO/Wappalyzer/blob/master/schema.json), additional properties of the emta object can only have strings as values. I also checked the [apps.json](https://github.com/AliasIO/Wappalyzer/blob/master/src/apps.json) and couldn't find any `generators` assigned an array value.

ScionOfBytes:
@hadifarnoud I've updated the docs.

gadcam:
```shell
Validating apps.json...
Validation Errors:

Instance is not a required type
uri: urn:uuid:633d6b5f-1dd8-4eb0-824d-630f23818429#/apps/Kamva/meta/generator
schemaUri: urn:uuid:6e98092c-8228-420a-bf5a-e095892ae426#/properties/apps/additionalProperties/properties/meta/additionalProperties

attribute: type
details: ["string"]
npm ERR! Test failed.  See above for more details.
```
@hadifarnoud I don't really know if the problem was formating or the thing you underlined... seems you're right.
In addition if you want to keep both patterns you can write it as it
```js
"meta":{
	"generator": "[CK]amva"
},

hadifarnoud:
@AliasIO any feedback?

AliasIO:
Merged, thanks all.

PR #1287, opened 2016-12-02
Update README.md

EvanGoss:
On my machine (OSX 10.12.1, Docker 1.12.3, Vagrant 1.9.0), I received the error below when running 'vagrant up' until I installed VirtualBox, even though I already had Docker installed.  Is it normal to need VirtualBox installed?

Error Message:
```
No usable default provider could be found for your system.

Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ which is managed by Vagrant.
 Running this environment is optional but recommended as it provides some helpful tools.
 
 First, install [Docker](https://www.docker.com/) and [Vagrant](https://www.vagrantup.com/)
-on your system.
+on your system. [VirtualBox](https://www.virtualbox.org/) might be needed as well.
 
 Clone the Wappalyzer repository and open the newly created directory in a
 terminal. Run `vagrant up` to start the environment.


ScionOfBytes:
@EvanGoss Vagrant is an open-source software product for building and maintaining portable virtual development environments. But it needs a backing vm engine. So for using Vagrant VirtualBox (or one of the other non-free providers) *have* to be installed. Vagrant, by itself, can't do anything.

EvanGoss:
@ScionOfBytes - thanks for responding. Can you elaborate on the README instructions to include the bare minimum of dependencies? I installed Docker and Vagrant and ran `docker pull wappalyzer/dev` but got the error mentioned in my comment above. I was thinking that Docker and VirtualBox were functionally similar and that having only Docker installed (w/o any other providers) would be sufficient.  I've only used VMWare and have no prior experience with VirtualBox, Vagrant or Docker, so I apologize for the gaps in my knowledge.  What am I missing?

ScionOfBytes:
@EvanGoss Okay, so I'm not exactly a virtualization expert, but here's what I know of the technologies.

VMWare, Parallels and VirtualBox are all virtualization providers that let you, well, virtualize OSs on top of whatever you're running. Vagrant is sort of the manager for these. Vagrant provides one single API for working with any of these Virtualization providers so you can think of learning about and working with just Vagrant (which has powerful API and configurations) and not any individual provider. But since Vagrant is just an abstraction on top of these, you still need one of these providers on your machine.

Docker, AFAIK, is an entirely different way to virtualize stuff and has a more layered-container model. This is totally separate from the world of Vagrant, VMWare, VirtualBox and Parallels although it tries to achieve the same thing.

So you can use Docker and Vagrant independently. If you've been using Docker and you get an error, that's an error related to Docker and nothing to do with Vagrant. 

Your first message said that when you ran Vagrant, you got a certain error. What did you get when you ran Docker?

Can you try the whole thing with only docker and let me know what happens? I'll try the build myself at some point and see if any instructions are missing or miscommunicated.

ScionOfBytes:
@EvanGoss Turns out you can also configure Docker to be a virtualization engine for Vagrant.
Have you installed Docker?

AliasIO:
The Vagrantfile sets the default provider to Docker so I'm not sure why it's complaining. You should be able to run `vagrant up --provider=docker` without having to install VirtualBox.

ScionOfBytes:
@AliasIO For me, when I tried this, it was a different error. Vagrant could detect my docker fine, but:

```
==> default: Docker host is required. One will be created if necessary...
    default: Vagrant will now create or start a local VM to act as the Docker
    default: host. You'll see the output of the `vagrant up` for this VM below.
    default:
    default: Box 'hashicorp/boot2docker' could not be found. Attempting to find and install...
    default: Box Provider: virtualbox
    default: Box Version: >= 0
The box 'hashicorp/boot2docker' could not be found or
could not be accessed in the remote catalog. If this is a private
box on HashiCorp's Atlas, please verify you're logged in via
`vagrant login`. Also, please double-check the name. The expanded
URL and error message are shown below:

URL: ["https://atlas.hashicorp.com/hashicorp/boot2docker"]
Error:
```

Any ideas what's going on here?

AliasIO:
Ah right, you need a Linux host to run Docker. Boot2Docker sets up a VirtualBox to facilitate that. There's also this but I've never used it (as a Linux user):

https://blog.docker.com/2016/03/docker-for-mac-windows-beta/

DaAwesomeP:
The Docker Hyper-V stuff is messy. After trying it, I got a BSOD at every boot and had to boot into safe mode to disable Hypervisor. Stick with the Docker VirtualBox method to be safe.

AliasIO:
It's working for me on macOS Sierra with Docker for Mac and Vagrant 1.9.2 (not 1.9.3). For some reason `vagrant up` prompts for the Docker user's password, which is `tcuser`.

http://stackoverflow.com/questions/31613483/vagrant-with-docker-provider-requires-rsync-password

AliasIO:
It took me most of a day but I got rid of Vagrant in favour of native Docker. Just run `./run links` to get started and `./run validate` to test changes before submitting a pull request. Tested on macOS Sierra, Docker version 17.03.1-ce.

PR #1289, opened 2016-12-02
Add YUI library regex with version

grmpyninja:
Moodle is using YUI and apparently it's not detected properly. Haven't tested the regex in Wappalyzer, because didn't have time to build the dev env, so please give it a try before merging.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -8213,8 +8213,11 @@
 			],
 			"env": "^YAHOO$",
 			"icon": "YUI.png",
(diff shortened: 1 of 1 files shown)


ScionOfBytes:
@grmpyninja Hey, could you build this locally and see if you get the tests to pass? No point going forward till the tests pass. The JSON seems to be invalid.

AliasIO:
Did you mean to include `xregexp` in the pattern? Can you show an example of a string that you're trying to match?

PR #1297, opened 2016-12-20
Klarna Checkout

vectart:
Klarna Checkout is a checkout and payment widget installed on more than 10,000 e-stores.
More details: https://www.klarna.com/uk/business/sell-online-with-klarna


AliasIO:
Thanks!

PR #1300, opened 2016-12-29
Adds new application Rakuten DBCore to apps.

manuwell:
Hi, 

I'd like to add this new application that properly match our applications. 
We are an e-commerce platform wrongly mapped as bigcommerce.

Thanks


AliasIO:
Thanks!

PR #1302, opened 2017-01-05
Aurelia JS framework detection

jhewt:
This adds a simple detection for the Aurelia, a Javascript Framework.
Since developers may use many different build tools the only way to detect it is to read attributes.


jhewt:
A site with a list of sites built with Aurelia to test the rules: http://builtwithaurelia.com/

src/apps.json
```diff
@@ -709,6 +709,21 @@
...
+		"Aurelia": {
+			"cats": [
+				12
+			],
+			"icon": "Aurelia.svg",
+			"html": [
+				"aurelia-app=[^>]",
+				"data-main=[^>]aurelia-bootstrapper",
+				"au-target-id=[^>]\\d"
+			],
```
AliasIO: The HTML patterns are too liberal and match plaintext. Please change it to something that contains a HTML opening tag, e.g. `<[^>]+aurelia-app=`.

jhewt: Indeed, fixed now. Thanks!

AliasIO:
Thanks!

PR #1311, opened 2017-01-18
detection for new wix.com behavior

wingsuitist:


ScionOfBytes:
@wingsuitist Hi. Could you please check why the test is failing and fix the PR? Thanks :)

src/apps.json
```diff
@@ -8065,6 +8065,7 @@
 				"Set-Cookie": "Domain=\\.wix\\.com",
 				"X-Wix-Dispatcher-Cache-Hit": ""
 			},
+			"html": "X-Wix-Renderer-Server",
```
AliasIO: This pattern is too liberal and should contain at least a HTML opening tag.

AliasIO:
Thanks!

PR #1315, opened 2017-01-23
added Foswiki detection

boossy:


ScionOfBytes:
@boossy Hi. Could you check why the tests are failing and make corrections?

src/apps.json
```diff
@@ -2483,6 +2483,15 @@
 			"script": "cartjs\\.php\\?(?:.*&)?s=[^&]*myfortune3cart\\.com",
 			"website": "fortune3.com"
 		},
+		"Foswiki": {
+			"cats": [
+				8
+			],
+			"html": "/foswiki/bin/../pub/System/CommentPlugin/comment\\.css",
```
gadcam: HTML patterns shouldn't match plain text see https://github.com/AliasIO/Wappalyzer/wiki/Specification
In order to correct it you could add some HTML tag in the pattern.

AliasIO:
Closing this one as per @gadcam's comment.

boossy:
I notice that Foswiki has been added to Wappalyzer, however, the results often point to sites that are not accessible, some because they require a login (which is ok), but others because they...
- are too slow? (ERR_CONNECTION_TIMED_OUT)
- simply don't exist... (ERR_NAME_NOT_RESOLVED)
It's strange that they appear in the list of Foswiki sites, isn't it?
Any ideas on how to get "valid" sites?

craiglondon:
@boossy maybe check out this URL http://foswiki.org/About/ExampleSites

PR #1317, opened 2017-01-24
Fix capital W in repo name in readme

digitalica:
simple minor issue i found getting started...

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Run `vagrant ssh` to access the environment and read usage instructions.
 
 ```shell
 $ git clone https://github.com/AliasIO/Wappalyzer.git wappalyzer
-$ cd wappalyzer
+$ cd Wappalyzer
 $ vagrant up
 $ vagrant ssh
 ```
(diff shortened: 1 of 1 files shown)


ScionOfBytes:
@digitalica Why should the `w` be upper-cased? This command `git clone https://github.com/AliasIO/Wappalyzer.git wappalyzer` clones it into a directory with lower-cased `w`.

ScionOfBytes:
@digitalica Yes, but why does it matter?

digitalica:
@ScionOfBytes Ah, you're right. I missed the new dirname in the  first line of the instructions, as I first created a fork on Github and then just cloned that, by copying the path from Github.

That gave me a directoryname with a capital W. (Ubuntu 16.04). I think that will be a common approach right?

So the instructions are correct as they are, its no important fix, but I think its a little better (less confusing) if the dirname matches the repo name...

AliasIO:
I'll leave it as-is, I personally prefer lowercase filenames. Also if you just leave the argument off it will create the folder with a capital W, no need to specify it in that case.

PR #1337, opened 2017-02-03
Fix #1325 - BigCommerce is wrongly detected

### LINKED ISSUE #1325: BigCommerce is wronly detected
Hello,

from what I can tell here:

https://github.com/AliasIO/Wappalyzer/blob/master/src/apps.json#L825

You are trying to detect a javascript object that will match the `$compareProducts^` regex and if it exists you conclude the customer is using BigCommerce. This is really shallow - in my code I have the javascript following function:

````
function compareProducts() {
````

and becase of that you think I use BigCommerce, which I'm **not**.

twindual:
Remove the "env" property that could incorrectly match a function in the global scope. Removed the trailing 'v' from the "script" and "html" entries that looks for the BigCommerce CDN paths since they can begin with letters other than 'v' like 's' in the to examples below.

Check against the following URLs :
https://shopmonarch.averydennison.com/
http://shop.panthers.com/

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -822,10 +822,9 @@
 			"cats": [
 				6
 			],
-			"env": "^compareProducts$",
-			"html": "<link href=[^>]+cdn\\d+\\.bigcommerce\\.com/v",
+			"html": "<link href=[^>]+cdn\\d+\\.bigcommerce\\.com/",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #1338, opened 2017-02-03
Fix anonymous reports option in Chrome

pd4d10:
Convert string `'0'` to number `0`, see https://github.com/AliasIO/Wappalyzer/pull/1050


AliasIO:
Thank you!

PR #1346, opened 2017-02-05
Improved Nette Framework detection

landsman:
Damn, Github closed my original PR #1344 after my changes. Here is it again.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -5043,9 +5043,9 @@
 		  },
 		  "env": "Nette",
 		  "html": [
-			"<(input|textarea)[^>]+data-nette-rules",
-			"<[A-z][^>]+id=\"snippet--",
-			"<(form|input)[^>]+id=\"frm-"
+			"<input[^>]+data-nette-rules",
+			"<div[^>]+id=\"snippet--",
+			"<input[^>]+id=\"frm-"
 		  ],
 		  "icon": "Nette Framework.png",
 		  "implies": "PHP",
--- a/src/apps.json
+++ b/src/apps.json
@@ -5034,15 +5034,22 @@
 			"website": "netsuite.com"
 		},
 		"Nette Framework": {
-			"cats": [
-				18
-			],
-			"headers": {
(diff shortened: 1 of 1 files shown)


src/apps.json
```diff
@@ -5034,15 +5034,22 @@
...
-			"cats": [
-				18
-			],
-			"headers": {
-				"X-Powered-By": "Nette Framework"
-			},
-			"icon": "Nette Framework.png",
-			"implies": "PHP",
-			"website": "nette.org"
+		  "cats": [
```
Majkl578: Wrong indentation (everywhere).

src/apps.json
```diff
@@ -5034,15 +5034,22 @@
...
+			18
+		  ],
+		  "headers": {
+			"X-Powered-By": "Nette Framework",
+			"Set-Cookie": "nette-browser="
+		  },
+		  "env": "Nette",
+		  "html": [
+			"<input[^>]+data-nette-rules",
+			"<div[^>]+id=\"snippet--",
```
Majkl578: Should only be one dash (two imply a presenter and would not work for components). And it might not be only div afaik (in case of `n:snippet`), but it's most common use case.

src/apps.json
```diff
@@ -5034,15 +5034,22 @@
...
+			"<input[^>]+id=\"frm-"
```
Majkl578: Should probably be checked against form as well.

src/apps.json
landsman: This is unique way I think

src/apps.json
landsman: fixed

src/apps.json
landsman: fixed

src/apps.json
landsman: fixed

landsman:
Is it right @AliasIO ?

AliasIO:
Looks good, thank you!

landsman:
Thank you for merge!

PR #1352, opened 2017-02-09
Added Roadiz CMS X-Powered-By header.

ambroisemaupate:
Added another parsing method for Roadiz CMS which can be recognized with `generator` tag and/or headers.


AliasIO:
Thanks!

PR #1362, opened 2017-02-11
Improve Bootstrap detection

aniirban101:
This PR improves upon bootstrap detection by including an edge case.

The earlier regex was unable to detect bootstrap if the `<div` was followed by a tab. This PR fixes this issue.


AliasIO:
Thanks!

PR #1370, opened 2017-02-19
Improve pattern HTML pattern to detect SolodevCMS markup

craiglondon:


src/apps.json
```diff
@@ -6921,7 +6921,7 @@
 			"headers": {
 				"solodev_session": ""
 			},
-			"html": "<div class='dynamicDiv' id='dd\\.\\d\\.\\d'>",
+			"html": "<div class=(?:\"|')dynamicDiv(?:\"|') id=(?:\"|')dd\\.\\d\\.\\d(?:\\.\\d)?(?:\"|')>",
```
AliasIO: These can be shortened to `[\"']`

AliasIO:
Thanks!

PR #1388, opened 2017-02-24
Fix phantomjs driver loading wappalyzer.js

RootLUG:
Fix for phantomjs driver to first look if wappalyzer.js exists in the current working directory before changing it to the driver's directory as it wasn't possible to run phantomjs driver directly from cloned repository without modifications due to missing files (e.g. wappalyzer.js doesn't exists by default in the driver directory).

Also logs errors to stderr if wappalyzer wasn't initialized due to injection error as driver would otherwise fail & crash

### DIFF
--- a/src/drivers/phantomjs/driver.js
+++ b/src/drivers/phantomjs/driver.js
@@ -10,10 +10,13 @@
 		quiet           = false; // Don't output errors
 
 	try {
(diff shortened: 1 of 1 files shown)


AliasIO:
The `wappalyzer links` command should be run before using the driver.

PR #1418, opened 2017-03-07
Fixed false EdgeCast CDN discovery

dmalshin:
Fixed: false discovery of EdgeCast CDN on Node's servers using github.com/jfhbrook/node-ecstatic middleware


AliasIO:
Nice one, thanks!

PR #1431, opened 2017-03-13
Fix WebExtension in Chrome-based Browsers

### LINKED ISSUE #1426: Firefox WebExtension
Firefox is ditching the old Addon SDK and replacing it with WebExtensions. The Chrome addon may already be compatible (I haven't tried it), but there is one change to it that would be beneficial. Sadly, displaying all of the icons in the URL bar will be no longer possible, but [Firefox places the `pageAction` buttons/popups in the URL bar](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/pageAction), so at least the Wappalyzer button will still have the option of not cluttering the toolbar.

DaAwesomeP:
- Simply doesn't use the polyfill for `onMessage` and `sendMessage()`
 - Add basic OS rules to `.gitignore` (I noticed when I was committing that it wanted to through in `Thumbs.db`)
 - Fix options page key in manifest (Firefox uses a different key and object that also works in Chrome)

I have tested this in Firefox 55 (Nightly), Chrome 56, Opera 45 (developer), and Opera 43. In case that anyone cares, no, this does not pass Edge's über-strict manifest rules.

Closes #1426


AliasIO:
Thanks. I'll try it out.

AliasIO:
That works, thanks a ton.

DaAwesomeP:
No problem! Wappalyzer will survive the _Great Addon Purge_.

AliasIO:
What needs to be done to make it work in Edge?

DaAwesomeP:
Edge is very very *very* ***very*** strict about the manifest file. It doesn't really have defaults to the keys; all keys that are needed must be present. Also, some of the keys have different schemas; I don't know why Microsoft does this. Also, I don't think that the polyfill really helps at all with it.

Opera, on the other hand, works almost exactly like Chrome (except that it puts the pageAction in the URL bar like Firefox). I don't think that there's a developer fee though, so the extension could definitely be published for Opera. I tested with Opera 43 (stable) and 45 (dev).

I'll mess around with Edge and try to get it to work.

DaAwesomeP:
First error in Edge:
```
SEC7111: HTTPS security is compromised by ms-browser-extension://Wappalyzer_b9gdh4rg3k5pe/js/inject.js
```
[Microsoft says to ignore the error](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8748330/#comment-2), but nothing else is in the log to show why it isn't working. There's simply no output anywhere. There is nothing in either the background page dev tools console or a tab's dev tools console. There is no pageAction in sight.

DaAwesomeP:
I've [pushed the manifest changes for Edge to my fork](https://github.com/DaAwesomeP/Wappalyzer/commit/138578af39fdd790034db3b64095abc1ce6b35ac) if you'd like to mess around with it. I hit the brick wall.

PR #1448, opened 2017-03-18
Use different WebExtension manifests per browser

DaAwesomeP:
I recently submitted an Opera addon, and they are very very picky about what's in the manifest (they don't want anything extraneous that doesn't apply to their browser).

Firefox
 - Supports SVG icons for main icon and page action
 - Needs minimum version 52 (see the red warning partly down [this page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Implement_a_settings_page))

Chrome/Opera
 - Doesn't support SVG icons in manifest
 - Can't have extra attributes in manifest

Edge
 - Uses other options page notation
 - Haven't tested SVG icons
 - Edge version still doesn't work


AliasIO:
Thanks, I'll have a play around with this. It may be nice to keep the Firefox and Chrome/Opera manifests the same and leave it as 'manifest.json' so it can be loaded as an unpacked extension.

DaAwesomeP:
Well, in order to submit to Opera at least they'll make you remove the `applications` key, you can't use SVG icons in Opera and Chrome, and the Opera mods will definitely catch having two options keys (for Edge).

AliasIO:
I suppose we can stick with PNG icons and work around the localStorage issue. We can have a separate file for Edge if required.

AliasIO:
I made some changes and submitted the extension to Opera. Let's see what they say.

PR #1451, opened 2017-03-20
Update Firebase Logo

sholomabrahams:
Firebase adopted this new logo when it upgraded to version 3 in May
2016.


AliasIO:
Thanks!

PR #1462, opened 2017-03-21
use absolute path to driver.js

bitmage:
When I install via NPM and run this as a library as described in the readme, I get an error saying that driver.js cannot be found.  This patch fixes the error by calculating an absolute path and passing that to phantomjs.


PumpUp:
I also get the same problem, and this patch is the good response

hbakhtiyor:
any reasons why the patch still not yet merged?

AliasIO:
Thanks!

PR #1467, opened 2017-03-22
Add detection for live chat Olark via script tag

craiglondon:
This can be tested here
http://tealium.com/integrations/


AliasIO:
Thanks!

PR #1468, opened 2017-03-22
ForkCMS -- Add detection for via meta tag

craiglondon:
Testable here
http://demo.fork-cms.com/en

And should work on any of the websites listed here
http://www.fork-cms.com/showcase


AliasIO:
Thanks!

PR #1470, opened 2017-03-22
BoldChat -- add detection via script

craiglondon:
Testable at
https://www.bask.com/
https://www.plannedparenthood.org/


AliasIO:
Thanks!

PR #1472, opened 2017-03-22
use absolute path

PumpUp:
in nodejs there is a path problem. This is a simple way to correct it.

### DIFF
--- a/src/drivers/phantomjs/index.js
+++ b/src/drivers/phantomjs/index.js
@@ -6,7 +6,7 @@ const
 	phantomjs = require('phantomjs-prebuilt');
 
 exports.run = function(args, callback) {
-	args.unshift.apply(args, ['driver.js', '--web-security=false', '--load-images=false', '--ignore-ssl-errors=yes', '--ssl-protocol=any']);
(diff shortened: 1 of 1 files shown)


src/drivers/phantomjs/index.js
```diff
@@ -6,7 +6,7 @@ const
 	phantomjs = require('phantomjs-prebuilt');
 
 exports.run = function(args, callback) {
-	args.unshift.apply(args, ['driver.js', '--web-security=false', '--load-images=false', '--ignore-ssl-errors=yes', '--ssl-protocol=any']);
+	args.unshift.apply(args, [__dirname+"/"+'driver.js', '--web-security=false', '--load-images=false', '--ignore-ssl-errors=yes', '--ssl-protocol=any']);
```
hbakhtiyor: better to use `path.join(__dirname, 'driver.js')`

craiglondon: I there there is another commit pending that is trying to solve this issue as well?

hbakhtiyor: @craiglondon not sure i understand

hbakhtiyor: yeah, exactly https://github.com/AliasIO/Wappalyzer/pull/1462/

PumpUp: #1462 ok, i close this request

PR #1478, opened 2017-03-26
WebExtension reference current window

### LINKED ISSUE #1465: WebExtension wrong popup
When clicking the page action in a second window, the popup of the tab selected in the first window shows up regardless of the tab of the second window. However, the icon of the page action of the second window tabs are correct. Are the inter-addon messages queuing about the state of each tab or about the state of a tab of a particular window?

DaAwesomeP:
Fixes #1465.

The tab query did not specify which window to select tabs from, so it would always return tabs from the first window, causing the popup to show incorrect data.


AliasIO:
Thanks!

PR #1495, opened 2017-04-02
Fix not answering any response if no apps detected

hbakhtiyor:
it hangs if not apps detected or can't detects apps.
e.g. for youtube.com


AliasIO:
Thank you!

PR #1500, opened 2017-04-04
Refactor Dockerfile

PeterDaveHello:
This change will make the docker image lighter(smaller) and also make the build process faster!
 - curl is builtin in phusion/baseimage, so there is no need to install it before nodesource
 - nodesource script will do apt-get update, so there is no need to do it manually
 - change security.ubuntu.com to use mirror site since it's very slow
 - separate apt package dependencies into different line so they can be easily traced by git
 - apt-get update cache clean up show be in the same layer of Docker image, or it'll not really work


AliasIO:
Nice one, thanks!

PR #1501, opened 2017-04-04
Move Travis CI to Ubuntu Trusty

PeterDaveHello:
As GCC v4.8 is already installed, there is no need to use ubuntu-toolchain-r-test ppa for it, then we can speed up every CI round!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "node"
 sudo: false
+dist: trusty
 before_install:
  - export WAPPALYZER_ROOT=$TRAVIS_BUILD_DIR
  - export WAPPALYZER_NODE_PATH=$TRAVIS_BUILD_DIR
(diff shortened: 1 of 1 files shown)


DaAwesomeP:
This actually seems to have run longer.
3min 34sec: https://travis-ci.org/AliasIO/Wappalyzer/builds/218532383
Average of 1min ~30sec: https://travis-ci.org/AliasIO/Wappalyzer/builds

Thos times aren't a good comparison though. The trusty builds are a beta, so that might have something to do with it. Travis also probably didn't use the `node_modules` cache from the older builds.

AliasIO:
What was it using before? Xenial?

PeterDaveHello:
precise @AliasIO

AliasIO:
Shouldn't we move to Xenial at least?

PeterDaveHello:
Travis CI haven't support Xenial yet.

AliasIO:
Thanks!

PR #1515, opened 2017-04-11
Option to keep icon from changing

### LINKED ISSUE #1166: Add preference to *not* change the icon
Maybe I'm the only one, but it bugs me that as I browse, the icon changes based on tech used on the site. I'd really rather just have one 'core' Wappalyzer icon in the UI. I'm finding myself being momentarily confused when I see a new icon there and wondering what extension that is - then I remember it's Wappalyzer changing on the fly.

DaAwesomeP:
Fixes #1166

When the box is unchecked, Wappalyzer will use the default purple icon; icons will not revert to the default until the page is reloaded again. Box is checked by default to change icon to first detected app.

I couldn't get the i18n to work. It just doesn't show the text.


AliasIO:
Great, thanks. I think all the i18n files will need to be updated, even if it's just the English string. I might also change it so the icon is grey by default and changes to purple when an application is detected.

PR #1522, opened 2017-04-13
add vinala and keep lighty

youssefhad:
Add the new name of Vinala and keeping Lighty

#1520


AliasIO:
Thanks!

youssefhad:
u r welcome :+1:

PR #1527, opened 2017-04-18
Improve detection of Froala Editor

stefanneculai:


AliasIO:
Thanks!

DaAwesomeP:
This commit is currently causing the build to fail: https://travis-ci.org/AliasIO/Wappalyzer/builds/224815468

gadcam:
Moreover it's not cautious to use the greedy <code>.*</code> that could, maybe, be replaced with a <code>[^"]*</code>.
@stefanneculai  what do you think about using <code>"html": "class=\\"[^\\"]*(?:fr-view|fr-box)"</code> instead ?

AliasIO:
Fixed in 20a41cbeadb4198e9de4d3593d8fe99a1fa8a215.

PR #1530, opened 2017-04-21
Adding Google Search Appliance match + icon

markbosky:
> 32x32 compressed (tinypng) .png logo from original GSA favicon .ico
> Applicable categories: web server (22),  search engine (29)
> Simple regex pattern matching via HTTP 'server:' header


AliasIO:
Thanks!

Issue #1516, opened 2017-04-11
add virtuoso detection


seralf:
Adding [virtuoso](https://virtuoso.openlinksw.com/) detection could be useful, since many linked open data projects used it. 
In many installation the default theme is used, than it's simple to detect its usage from header and footer, however other way to recognize it could be looking for paths like `/fct/{something}.vsp`

jvoisin:
Do you have a list of websites where this framework/application/… is deployed?

seralf:
there are many, why the list?

2017-04-20 16:12 GMT+02:00 jvoisin <notifications@github.com>:

jvoisin:
So I can implement its detection and tests it.

seralf:
Hi I've added a first draft for Virtuoso detection.

Those are examples of endpoints provided with Virtuoso:
+ http://dbpedia.org/sparql
+ http://live.dbpedia.org/sparql
+ http://bio2rdf.org/sparql
+ http://lod.openlinksw.com/sparql

Actually there could be other pages on the sites as well as the sites themselves hosted on Virtuoso (for example: http://lod.openlinksw.com/), but it's not simple to detect them. The endpoint is simpler to analyze since it usually has a standard URL, but on the other hand not all the SPARQL endpoint are hosted on Virtuoso.

PR #1532, opened 2017-04-21
added Virtuoso detection

seralf:
added Virtuoso detection, as requested in #1516

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -8442,7 +8442,9 @@
 				"Copyright": "^Copyright &copy; \\d{4} OpenLink Software"
 			},
 			"url": "^https?://(?:[^\\/]+\\.)?.*?/sparql\\;confidence:20",
-			"Server": "Virtuoso:*?\\;version:\\1"
+			"headers": {
+				"Server": "^.*Virtuoso\/(\d{2}.\d{2}.\d{4}).*$;version:\\1"
+			}
 		},
 		"Visual WebGUI": {
 			"cats": [
--- a/src/apps.json
+++ b/src/apps.json
@@ -8441,7 +8441,7 @@
 				"Keywords": "^OpenLink Virtuoso Sparql",
 				"Copyright": "^Copyright &copy; \\d{4} OpenLink Software"
 			},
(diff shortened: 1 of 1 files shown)


AliasIO:
There is no "Server" property, you probably meant to use "headers". Also you're using a backreference for the version number without using a capture group.

seralf:
hi thanks for the suggestion. I've updated the configuration for Virtuoso, is it correct now? (I still see travis failed but never understand why it's failing)

seralf:
wow, I think I have collected far more errors than anyone else, but it finally seems to be ok :-)

src/apps.json
```diff
@@ -8433,6 +8433,20 @@
...
+		"Virtuoso": {
+			"cats": [
+				"34"
+			],
+			"website": "https://virtuoso.openlinksw.com/",
+			"meta": {
+				"Keywords": "^OpenLink Virtuoso Sparql",
+				"Copyright": "^Copyright &copy; \\d{4} OpenLink Software"
+			},
+			"url": "(?:)[^\/]+\/sparql\\;confidence:20",
```
gadcam: This pattern isn't relevant as it will trigger a lot of false positive IMO. What do you think @AliasIO  ?

seralf:
what pattern? do you mean the url?
Is relevant as we expect almost every virtuoso installtion to have a
/sparql path, but it can be found for every triplestore as well. My idea
was to provide an enforcement to other rules, if you think it is not needed
please remove it: I have no way to test it directly

2017-05-19 18:41 GMT+02:00 Camille Barneaud <notifications@github.com>:

gadcam:
If you open the pull request in github you will see I commented under the URL pattern ;)
I'm only a contributor so I can't remove anything and take my comments for what they are : advises.
IMO you don't need this pattern as it will trigger more false positive than good detections and you already found reliables ways to detect Virtuoso.

src/apps.json
AliasIO: Agree, it should be removed or made more specific.

seralf:
ok, I added all the suggested modifications

PR #1535, opened 2017-04-25
Add SweetAlert2 app

limonte:
https://github.com/limonte/sweetalert2

Related to https://github.com/limonte/sweetalert2/issues/388

Currently Wappalyzer detects SweetAlert plugin here:

and that's incorrect, it should be SweetAlert**2** - those two projects are different.


AliasIO:
Thanks!

limonte:
Thanks for merging! When will be the next deploy to https://wappalyzer.com/ so I can double-check if this change is working correctly?

limonte:
@AliasIO ping :)

AliasIO:
Should this not be merged with the existing SweetAlert app?

limonte:
SweetAlert and SweetAlert2 are 2 different projects, so it's okay to have both of them.

My question is - why it still doesn't show SweetAlert2 for this page: https://limonte.github.io/sweetalert2/

AliasIO:
That's because it uses the NPM module which hasn't been updated yet.

AliasIO:
It's there now but it detects both SweetAlert and SweetAlert2. The latter should have `"excludes": "SweetAlert"` to avoid this.

limonte:
Thanks for updating the NPM module for http://wappalyzer.com

I just added `"excludes": "SweetAlert"` to the SweetAlert2 configuration as you suggested.

Thanks for the support! :beers:

PR #1523, opened 2017-04-13
Updated PHP icon (svg)

DmitryKrekota:
I used the image from: https://en.wikipedia.org/wiki/PHP#/media/File:PHP-logo.svg
Permission details: "This work is free and may be used by anyone for any purpose".

PR #1536, opened 2017-04-25
Make PHP.svg square icon - 400x400

DmitryKrekota:
Fix not square sizes for PHP.svg icon. The problem was described by @AlexByte in #1523 PR.


AliasIO:
Thanks!

PR #1537, opened 2017-04-27
Add Fastly detection

ezkl:
I've added detection for [Fastly](https://www.fastly.com/). The existence of a `Fastly-Debug-Digest` header isn't a perfect indicator, but it shouldn't result in false positives.


AliasIO:
Thanks!

PR #1538, opened 2017-04-28
Add Raspbian OS detection

powell759:
I've added detection for [Raspbian](https://www.raspbian.org/).

There are a few changes in app.json that aren't relevant. These are just from running the validator. I don't see any reason to unfix them. Not sure why it wasn't already validated.


AliasIO:
Thanks!

PR #1539, opened 2017-04-28
Add Storyblok detection.

DominikAngerer:
I've added detection for [Storyblok](https://www.storyblok.com).

We're a headless CMS so we won't force someone into using the generator meta tag - but it will be the only way to track those pages.

* https://www.wallsplashapp.com/
* ...


AliasIO:
Thanks!

PR #1548, opened 2017-05-03
Add Play Framework

nremond:


ctl:
Isn't icon is too big?

nremond:
Andrey, would a SVG be better? Or, what would be an ideal PNG height/width?

ctl:
@nremond , i think that SVG is a best choice , but i didn't find any SVG logo of Play Framework.

DaAwesomeP:
SVGs are preferred to PNGs. PNGs should be 32x32 or at least 32px in one dimension if the image isn't square. PNGs should also be cropped to the smallest size (no unnecessary padding). They are losslessly compressed during build, so don't worry about the file size.
https://github.com/AliasIO/Wappalyzer/wiki/Contributing#adding-a-new-application

nremond:
Thanks for the feedback. I switched the SVG logo.

ctl:
uhh... I don't think that "Play" text is necessary

nremond:
Ok, I'll try to find the logo only.

ctl:
@nremond uhh... https://github.com/AliasIO/Wappalyzer/pull/1548#issuecomment-299479422

nremond:
It found the SVG logo with only the icon and rebased my commits.

ctl:
cool

AlexByte:
Make the icon square.

nremond:
Done

AlexByte:
You could just set the same width and height for the svg tag.

nremond:
Thats what I did:
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 160 265" enable-background="new 0 0 160 265">`
=>
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 265 265" enable-background="new 0 0 265 265" preserveAspectRatio="xMidYMid meet">`

AlexByte:
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 160 265" enable-background="new 0 0 160 265"><g>`
=>
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 265 265" enable-background="new 0 0 265 265" preserveAspectRatio="xMidYMid meet"><g transform="translate(52.5)">`

DaAwesomeP:
@AlexByte why does the icon need to be square? The original icon was just fine. As a vector, it will resize proportionally to fit the space provided. I would not recommend editing the SVG unless to remove junk elements or blankspace/padding around the main logo/icon.

AlexByte:
I was wrong, the SVG icon does not have to be made square, the viewBox attribute must be set. https://github.com/AliasIO/Wappalyzer/issues/1564#issuecomment-300029642
The last commit can be removed.

nremond:
I removed the last commit.

AliasIO:
Thanks!

szepeviktor:
@nremond Could you add `PLAY_LANG` cookie also?

PR #1554, opened 2017-05-05
Add support for detection of govCMS (Drupal distribution).

seanhamlin:


AlexByte:
Drupal already implies PHP.

seanhamlin:
Pushed a commit that removes PHP as being implied.

AliasIO:
Thanks!

PR #1556, opened 2017-05-05
Add minified version of translucide js

simonvdfr:
Add minified version of translucide js detection


AliasIO:
Thanks! I resolved the merge conflict locally.

PR #1561, opened 2017-05-08
Fix Node.JS category

AlexByte:


AliasIO:
I think Node.js makes more sense in the Programming Languages category than Web Frameworks. Express is a framework, which runs on Node.js. We could change it to JavaScript but that can be confused with client-side JavaScript.

PR #1571, opened 2017-05-15
CloudCart CDN recognition added

pilievwm:
Hello, for some reasons CloudCart app is not recognizable anymore. I`ve added a new way of recognition. All static files are coming from our CND domain cdncloudcart.com
I hope that this change will fix the issue


pilievwm:
It is strange conflict. It seems that is missing curly bracket at the end.

AlexByte:
This is not a strange conflict, before adding the modifications, you need to update the data of the remote repository and reset the local repository data to match master branch of upstream repository.
```
git fetch upstream
git reset --hard upstream
```
Or for noobs
`git pull upstream`
Modifications are best done in a separate branch.

AlexByte:
Did you test this?
`"script": "^cdncloudcart$",`
Why did you delete it?
`"script": "/cloudcart-(?:assets|storage)/",`

AliasIO:
I fixed a bug that prevented meta tags from being checked, which fixes the CLoudCart issue.

pilievwm:
Guys the problem is that CloudCart app is not recognizable from Wappalyzer and it is not suggested as the main app. Plase check the screen
image

AliasIO:
That's related to #1588.

PR #1579, opened 2017-05-19
added RDF/SPARQL languages

seralf:
added RDF/SPARQL languages, for capturing dependencies in sites exposing SPARQL endpoints or JSON-LD API

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -6588,6 +6588,12 @@
 			"icon": "Raspbian.svg",
 			"website": "https://www.raspbian.org/"
 		},
+		"Rdf": {
+			"cats": [
+				"27"
+			],
+			"website": "https://www.w3.org/RDF/"
+		},
 		"React": {
 			"cats": [
 				"12"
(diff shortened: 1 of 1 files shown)


seralf:
NOTE: once the pull will be merged, it will be possible to add the "implies" on the related servers (for example CKAN should imply RDF, and Virtuoso should imply RDF/SPARQL)

gadcam:
You need to add an icon for each technology see [Adding a new app](https://github.com/AliasIO/Wappalyzer/wiki/Contributing#adding-a-new-application) in the wiki.
@AliasIO the build should have failed or I'm wrong ?

seralf:
uhm for RDF there is an icon, but SPARQL does not have a proper icon, I
suppose we could use for it the same icon as RDF, however

2017-05-19 18:49 GMT+02:00 Camille Barneaud <notifications@github.com>:

gadcam:
I think using the same icon is a good option as SPARQL always come with RDF.
Shouldn't SPARQL implies RDF ?

AliasIO:
@gadcam The icon field is no longer required, it falls back to `default.svg`.

gadcam:
@AliasIO Sorry I misssed this change. Shouldn't it raise a warning so the build won't fail but newcomers would be aware that they could use an icon ?

PR #1584, opened 2017-05-22
i have updated as per request.

MicroDreamIT:
Please take a look sir.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -7154,12 +7154,18 @@
 				"18"
 			],
 			"env": "^Shapecss$",
-			"html": "<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+			"html": [
+				"<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+				"shapecss\\.(?:min\\.)?(?:[a-f\d]+\\.)?css",
+				"shapecss.*\\.css"
+			]
 			"icon": "Shapecss.svg",
 			"script": [
 				"shapecss[-.]([\\d.]*\\d)[^/]*\\.js\\;version:\\1",
 				"/([\\d.]+)/shapecss(?:\\.min)?\\.js\\;version:\\1",
-				"shapecss.*\\.js"
+				"shapecss.*\\.js",
+				"shapecss\\.(?:min\\.)?(?:[a-f\d]+\\.)?js",
+				"<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.js",
 			],
 			"website": "https://shapecss.com"
 		},


MicroDreamIT:
I have done as per request, what should i do now?

src/apps.json
```diff
@@ -7154,12 +7154,18 @@
 				"18"
 			],
 			"env": "^Shapecss$",
-			"html": "<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+			"html": [
+				"<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+				"shapecss\\.(?:min\\.)?(?:[a-f\d]+\\.)?css",
+				"shapecss.*\\.css"
+			]
```
nekosune: I believe a ```,``` is needed here?

AlexByte: @nekosune What makes you think that

MicroDreamIT: no idea :O

AlexByte:
What request?

src/apps.json
```diff
@@ -7154,12 +7154,18 @@
 				"18"
 			],
 			"env": "^Shapecss$",
-			"html": "<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+			"html": [
+				"<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.css",
+				"shapecss\\.(?:min\\.)?(?:[a-f\d]+\\.)?css",
```
AlexByte: You should write `[a-f\\d]`.

src/apps.json
```diff
@@ -7154,12 +7154,18 @@
...
+				"shapecss.*\\.css"
```
AlexByte: Where is link tag?

src/apps.json
```diff
@@ -7154,12 +7154,18 @@
...
+			]
 			"icon": "Shapecss.svg",
 			"script": [
 				"shapecss[-.]([\\d.]*\\d)[^/]*\\.js\\;version:\\1",
 				"/([\\d.]+)/shapecss(?:\\.min)?\\.js\\;version:\\1",
-				"shapecss.*\\.js"
+				"shapecss.*\\.js",
+				"shapecss\\.(?:min\\.)?(?:[a-f\d]+\\.)?js",
+				"<link[^>]* href=\"[^\"]*shapecss(?:\\.min)?\\.js",
```
AlexByte: You should not use link tag. Еhe comma at the end of the row redundant.

src/apps.json
AlexByte: test

AlexByte:
Provide a list of sites other than the official one that use this technology.

src/apps.json
nekosune: To make it valid Json, like https://github.com/AliasIO/Wappalyzer/blob/master/src/apps.json#L271
As this is an object by the name of html inside another object alongside other elements, as per the Json specs, as found here, it needs a comma. http://www.json.org

AlexByte: I agree. All errors 
image

AliasIO:
Still broken, closing.

PR #1585, opened 2017-05-24
Added Navegg

wmarquardt:
Hi,

I've added support for Navegg, an Analytics platform from Brazil. (http://www.navegg.com)


AliasIO:
Thanks!

PR #1593, opened 2017-06-03
Fix Meteor implies

branchard:
Meteor does not imply jQuery if you use `static-html` package instead of `blaze-html-templates` package in case you are using `React` for [example](https://guide.meteor.com/react.html#using-with-meteor).


AliasIO:
Thanks!

PR #1626, opened 2017-06-29
add class detection to UIkit

bbashy:
Adds `uk-container` and `uk-section` to detection.

_Also update website link to https_


AliasIO:
Thanks!

PR #1630, opened 2017-07-05
update adnegah

foadabdollahi:
i wanna remove non-related website from https://wappalyzer.com/applications/adnegah.
 
1 stackoverflow.com
2 askubuntu.com
3 api.jquery.com
4 magento.stackexchange.com
5 wordpress.stackexchange.com
6 serverfault.com
7 superuser.com
8 unix.stackexchange.com
9 drupal.stackexchange.com
10 jqueryui.com

these site do not use adnegah service but show adnegah icon in wapplayer!

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -254,12 +254,14 @@
 		"Adnegah": {
 			"cats": [
 				"36"
-			],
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #1632, opened 2017-07-09
Use arrays for "cats" values

pshevtsov:
Use arrays for "cats" values for imperia CMS and pirobase CMS


AliasIO:
I fixed this one already, thanks.

PR #1635, opened 2017-07-10
fixed missing quotes on Category for uCoz

dahawk:
was discovered because one of our build tests started failing


AliasIO:
Thanks!

PR #1642, opened 2017-07-11
Fix apps.json schema validation

EvaSDK:
Basically the same as PR #1632, #1633 and #1635.
Re-order file using jq while at it. This should probably make it into a test ran by Travis so it doesn't get broken so often.


AliasIO:
Thanks, these have been fixed.

PR #1644, opened 2017-07-14
Added Contentful

cachrisman:


src/apps.json
```diff
@@ -1646,6 +1646,19 @@
 			},
 			"website": "http://www.gocontentbox.org"
 		},
+		"Contentful": {
+			"cats": [
+				"1"
+			],
+			"html": [
+				"<[^>]+(?:https?:)?\/\/images.contentful.com",
```
gadcam: You don't need to escape slashes and you need to escape dots.
Moreover you could unify all theses patterns in this one:
<code><[^>]+(?:https?:)?//(?:assets|downloads|images|videos)\\\\.contentful\\\\.com</code>

cachrisman: updated to reflect your suggestion

AliasIO:
Thanks!

PR #1661, opened 2017-07-30

AndreiMaz:
1. Added Generator Meta

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10069,9 +10069,15 @@
 			"cats": [
 				"6"
 			],
+			"headers": {
+				"Set-Cookie": ".Nop.Customer|Nop.customer",
+			},
 			"html": "(?:<!--Powered by nopCommerce|Powered by: <a[^>]+nopcommerce)",
 			"icon": "nopCommerce.png",
 			"implies": "Microsoft ASP.NET",
+      "meta": {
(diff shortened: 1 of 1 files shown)


src/apps.json
```diff
@@ -10069,9 +10069,15 @@
 			"cats": [
 				"6"
 			],
+			"headers": {
+				"Set-Cookie": ".Nop.Customer|Nop.customer",
```
AliasIO: Shouldn't this just be `"Nop\\.customer"`?

AndreiMaz: @AliasIO Hi Elbert. Bot sure about how it should be written. Previous versions used "Nop.customer" and new versions use ".Nop.Customer" (starts with a dot). Please assist on how it should be better written

gadcam: I think the better would be <code>Nop\\\\.customer=</code>.
The dot need to be escaped so it will only match a dot and adding the equal sign will ensure that it is the name of the cookie not its content.
<code>Nop\\\\.customer=</code> is in <code>\\\\.Nop\\\\.customer=</code> so it will also match.
(Please note that the patterns are case insensitive)

AndreiMaz:
@AliasIO Thanks again. I've just a new pull request https://github.com/AliasIO/Wappalyzer/pull/1666

PR #1663, opened 2017-07-30
message.tracking_enabled wrapped by try{}catch(){}

geff21st:
image
Prevents error in console.


geff21st:
If you will approve this simple pull request, i'll stop seeing this annoying error in console every single time i'm trying to debug something..

AliasIO:
It's simpler to do `message && message.tracking_enabled`, I'll add that in.

PR #1665, opened 2017-07-31
Fix syntax issues in wappalyzer.js

fpoumian:
While working on https://github.com/AliasIO/Wappalyzer/pull/1652 I noticed I started having some issues testing the extension locally with Mozilla Firefox 55, namely that I could not see the extension icon anywhere. I was able to see it again after I fixed some syntax issues introduced in f61697c9f4a87574dd5ef4f462fe2d38b3c9237e in the wappalyzer.js file.


AliasIO:
Thanks a bunch!

Issue #1226, opened 2016-08-12
Enable detection using a JavaScript function


gadcam:
What about adding a new field named something like "test" where you can add a function that would return True when the application is present and False otherwise ? (or a version number..)
See https://github.com/johnmichel/Library-Detector-for-Chrome/blob/master/library/libraries.js.
What do you think @AliasIO ?

AliasIO:
I'm not a fan of the idea as it can interfere with JavaScript on web pages. Wappalyzer does inject a script but it's very minimal, just to read variables in the window object.

simonbrunel:
That would definitely help to determine the exact version number of a library loaded, for example, from a CDN that allows version range (https://cdn.jsdelivr.net/npm/chart.js@2/dist/Chart.js). Here, the detected version is `2` but it serves version `2.6.0`.

Though, I totally agree that running (potentially random) code inside web pages is not ideal for performance, neither for security. I really don't know what would be a better alternative (I'm sure you already spent time to figure out), maybe `env` could accept an extra parameter to test the string representation of the value of matching variables:

```javascript
// global variable
chartjs_version = '2.6.0';   
```

```json
// apps.json
"env": "^chartjs_version$\\;value:^([\\d.]+)$\\;version:\\1"
```

PR #1671, opened 2017-07-31
Add Chart.js

simonbrunel:
`Chart` seems a pretty common variable name (same for `Chart.js` script filename), please let me know if you think of a better way (than `confidence`) to detect this library.


gadcam:
First, I think you could add a pattern to detect the CDN version of Chart.js with version detection. (see https://cdnjs.com/libraries/Chart.js)
Then you could find some CSS selectors related to Chart.js, if there are some, or maybe some other global variables specific to Chart.js but I didn't find something obvious.

simonbrunel:
@gadcam thanks for the CDN suggestion. We don't inject CSS right now and I don't think of any other global variables that could help.

simonbrunel:
Thanks @jvoisin. I'm wondering if there is a way to retrieve the version number from a global variable?

gadcam:
Currently, the only way would be to extract it from the name of the global variable.

simonbrunel:
Ok, nevermind, was just looking for a solution for scripts like:

[`https://cdn.jsdelivr.net/npm/chart.js@2/dist/Chart.js`](https://cdn.jsdelivr.net/npm/chart.js@2/dist/Chart.js)

to be more precise. Right now the detected version will be `2` but the CDN serves `2.6.0` (same for [`@latest`](https://cdn.jsdelivr.net/npm/chart.js@2/dist/Chart.js))

gadcam:
I understand what you mean, see #1226 and #277.
We could detect a lot more information with something like that : however is that possible ? What will be the drawbacks of this, especially regarding performance ? If you want to discuss about that you are free to open an issue.

simonbrunel:
I think it's fine for now: I left a comment on #1226, if something moves on that side, we can still submit a new PR with a better version detection.

AliasIO:
Thanks!

PR #1672, opened 2017-08-01

AndreiMaz:
1. Added Generator Meta


AliasIO:
Thanks!

PR #1688, opened 2017-08-17
Update schema.json

FlashYo:
According to https://github.com/AliasIO/Wappalyzer/wiki/Specification, the filed "url" in schema.json have error.


AliasIO:
I fixed the wiki, thanks. ;)

PR #1695, opened 2017-08-27
Adding ability to use an optional headless browser

tylerbrownhenry:
Hi guys, I was sort of changing your code to use this with headless chrome, since I already had the data, and I just wanted to plug this into it, rather than making a second http request to get the same data.

This seems to test fairly well.  I think the worse thing is my documentation ,since my headers are actually Map objects rather than normal objects, so I wasn't sure how to document it.  But I wrote up the gist of how I was using it.

Anyway, if you find it useful, feel free to use it, it seemed like something worth sharing.

Great project you have saved me a ton of time on my project.

### DIFF
--- a/src/drivers/npm/README.md
+++ b/src/drivers/npm/README.md
@@ -38,3 +38,36 @@ wappalyzer.analyze('https://wappalyzer.com')
     console.error(error);
   });
 ```
+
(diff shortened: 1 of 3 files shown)


AliasIO:
Thanks! I might revisit this later, I'll leave the PR open for now.

AliasIO:
I'm closing this as the NPM driver has evolved quite a bit and a headless Chrome implementation should probably just be its own driver. Drivers are intended to be an interface between Wappalyzer core and a platform.

PR #1696, opened 2017-08-27
Update Typecho Checking

kokororin:
Typecho has a global variable `TypechoComment` and it has a meta tag like WordPress.


AliasIO:
Thanks!

PR #1699, opened 2017-08-27
Add ef.js & Neonclear w/ Chinese translation

ClassicOldSong:
Add Chinese translation with ef.js/ef-core detection


AliasIO:
Great, thank you!

PR #1680, opened 2017-08-11
Add Zeit's now & next.js.

fabe:
Implements the detection of [now](https://zeit.co/now) and [Next.js](https://zeit.co/next).

PR #1702, opened 2017-08-27
Remove too broad rule for Next.js

fabe:
As discussed in #1680 and #1706, html rule is too broad. The other two rules should cover the detection sufficiently.


AliasIO:
Thanks!

PR #1709, opened 2017-08-28
Reject if response property is null

vaidas-lungis:
Else you will face an Error 
> Can not access property of null


AliasIO:
Thanks!

PR #1721, opened 2017-09-03
Remove unnecessary popup body height attribute

AlexByte:


AliasIO:
Thanks.

AliasIO:
I'm reverting this one. Firefox needs the height attribute.

AlexByte:
What does this lack of attribute lead to?

PR #1749, opened 2017-09-28
Improve React detection

kirainmoe:
React 16 has removed ```data-reactroot``` attribute and wappalyzer can not detect React 16 correctly.
But React has a global variable ```__REACT_DEVTOOLS_GLOBAL_HOOK__```.


AlexByte:
LOL, this variable is set by extension.

PR #1752, opened 2017-10-01
Added Bulma

neoneosa:
I just added Bulma to the detection list. Not all sites will be detected as it uses the normal HTML filter, looking for a bulma(.min).css file. Tested in Opera.


AliasIO:
Thanks!

ruchern:
I do not get to see my page being detected as Bulma even with this addition. I am using Webpack and Bulma is being compressed into a generic `app.css` file.

gadcam:
Can you give us an example page to check what is possible to do ?
However I do not have much hope it is ***packed*** and with a ***generic*** name.

itsacorn:
I didnt know what to check for, so i just added a simple file name detection thing. I was gonna do some thing to check if specific classes exist, but some other frameworks might use the same name. I would rather not detect all sites than to get false positives.If someone knows a better way of detecting Bulma, I would appreciate contributions.

PR #1774, opened 2017-10-19
Adds Color Mautic Icon

robwent:
Was MIA when I put in the first submit.
This one might be more recognisable by people who already use it.


AliasIO:
Thanks!

PR #1777, opened 2017-10-25
Add CoinHive and CoinHive Captcha

q--:


AliasIO:
Thanks!

q--:
Hmm, I see a lot of popular websites in the  [list](https://wappalyzer.com/applications/coinhive) - Twitter, Reddit, VK, YouTube etc. I guess there's people with a hijacked web browser polluting the results.

I wonder if there's a way we could filter those out?

jvoisin:
Or maybe those are false-positives :/

AliasIO:
It certainly looks like someone is sending fake data. I made a change to try and combat this but if you have any ideas I'd be interested to hear them.

q--:
@jvoisin Could be, but I don't see any obvious errors in my code. Stories like [this one](https://www.theregister.co.uk/2017/10/23/cryptocurrency_miner_google_chrome_extension/) make me suspect that it's browser extensions injecting CoinHive into web pages.

@AliasIO I suppose you could make a list of websites which you know don't have CoinHive on them and if it gets reported on that website anyway, ignore CoinHive reports from that IP address.

Of course that's not bulletproof, because it won't remove CoinHive reports that were sent before the IP got blacklisted, people using multiple IP addresses etc. And IP addresses could get reassigned to somebody else who isn't infected.

And then there's the risk that a website you 'know' won't have CoinHive on it gets hacked to serve the script.

Assigning each Wappalyzer install an unique identifier and sending it with the reports could take care of the "IP addresses get reassigned" problem if the false reports aren't malicious, and temporary storing the identifier with reports so you can remove old reports of CoinHive if a client sends in a known false detection could take care of the other problems - but with the obvious privacy implications, and the overhead of storing the extra data.

But if my suspicion is correct and the CoinHive script gets injected into *every* web page, that means it will also be injected in the "Wappalyzer has been upgraded" page. If people haven't disabled the option to show that page, you would probably be able to detect at least some of the false reports early.

jvoisin:
haha, people are amazing, rogue extensions that are mining stuff :')

gadcam:
I agree with @q--  If it is possible it could be good to make a test page like https://wappalyzer.com/test where there is nothing to detect and report on each install/update if there is something strange.

AliasIO:
The issue isn't just with CoinHive, although #1807 is interesting. I noticed that Twitter and YouTube for instance shows up on dozens of applications, originating from many different IP addresses.

Maybe we should take this conversation off-line. What's the easiest way to do this, email perhaps?

gadcam:
Email is good for me but we could maybe create a slack project for this sort of situation.

jvoisin:
Cool kids aren't using IRc anymore?

AliasIO:
Ah yeah, Slack. Brilliant. Inviting @gadcam, @jvoisin, @q--, @DaAwesomeP, @AlexByte and @craiglondon. I think I'll accept anyone but I'll create a channel just for us.

https://join.slack.com/t/wappalyzer/shared_invite/enQtMjY4Nzc2MjczMjUxLTk4NTFmODBiNDQ4MGYyNTE5ODU2Zjg1N2Y5YmM1NzdkZTUyYjQ1MjE1OGU1NzUxZjY4ZGFlOWQxYWQ1ZDQwNjQ 

@jvoisin I'm sure there is an IRC plugin for Slack? :)

DaAwesomeP:
@jvoisin @AliasIO You can connect IRC/XMPP in the Slack account settings:
slack

AliasIO:
I noticed that, it's enabled. Looks like @jvoisin has already come to terms with modern times though.

jvoisin:
It's ok, I'll move to IRC ASAP ♥

PR #1790, opened 2017-10-31
Add Fingerprintjs

q--:
Library for browser fingerprinting.

Gave it the category "JavaScript frameworks", but maybe "Miscellaneous" would be better.


AliasIO:
Thanks!

PR #1797, opened 2017-11-05
Wordpress Theme & Plugin Detection

motealle:
Word-press Theme & Plugin Detection would be so nice because, WP is used on 23% of websites on the globe running a cms and some folks including me wanna find some details about sites running a Word-Press installation. (I often use www.whatwpthemeisthat.com ) and it would be nice if it was available on Wappalyzer plugin itself.


AliasIO:
I wouldn't mind having a look at your code bit this PR contains 8.5K changes. Can you please rebase?

PR #1799, opened 2017-11-05
PHPoole: replace logo by the black version.

Narno:
@AliasIO was right!
> Thanks. Do note that the PNG will be invisible for most people as it's white on white.
-- https://github.com/AliasIO/Wappalyzer/pull/1775#issuecomment-341624003


AliasIO:
Cheers!

PR #1809, opened 2017-11-07
Update apps.json, add MineHoster with svg icon

HAlexTM:


AliasIO:
Please see https://wappalyzer.com/docs#new-application:

> Only widely used applications should be added. When creating a pull request, please include ten or more links to websites that use the application, a GitHub page with at least 1,000 stars or anything that will help establish the size of the user base.

HAlexTM:
Ok thank you, I didn't read that, the last time it was approved without any
problem.

AliasIO:
New policy! Thanks for understanding.

PR #1818, opened 2017-11-10
Add Word and Excel detection

q--:


AliasIO:
Sweet, can't wait to see which websites are powered by Excel.

PR #1830, opened 2017-11-13
Added German Language

dabenzel:
I've added the german language files.
Some Strings can't really be translated, because they're identical in german or the translation would be strange.


AliasIO:
Nice one, thank you.

PR #1831, opened 2017-11-14
Low confidence for 2-character environment variables

q--:
These can be automatically generated by the Google Closure Compiler (which will, if configured, replace global variables with auto-generated ones such as `aa`, `ab`, `ac` et cetera.) Hence, they are a less reliable
indication that a library which would also set said variable is present.


AliasIO:
Thanks!

PR #1832, opened 2017-11-14
Changed german translations

dabenzel:
Yesterday I may have missed some. Also I did some rethinking on better translations.


AliasIO:
Thanks!

PR #1836, opened 2017-11-15
Fixed userAgent param doing nothing for npm driver

avasilkov:
If you run npm driver with --userAgent specified, it's not used and on request to the server used the default value of "Mozilla/5.0 Chrome/10.0.613.0 Safari/534.15 Zombie.js/5.0.7"

Moved userAgent init to the line of Browser() instance creation instead of require


AliasIO:
Thanks!

PR #1842, opened 2017-11-16
Add Google Closure Compiler

q--:
The Google Closure Compiler only sometimes sets variables named like this.


q--:
I just became aware of the fact that the Google Closure compiler is [part of a bigger toolset](https://en.wikipedia.org/wiki/Google_Closure_Tools), so now I'm not sure anymore if the variables I'm detecting are actually part of the Closure Compiler or if they're set by one of the other _Closure_ tools, such as the Closure Library (which is a prime suspect for me).

For reference, if anyone else wants to give figuring this out a shot, the variable names I'm seeing are:
```
    closure_uid_<number>
```
and

    closure_lm_<number>

e.g.

    closure_uid_579667508
    closure_lm_262899

These can be found at e.g. [Google Maps](https://maps.google.com)

I haven't managed to have [the Closure Compiler] generate JavaScript that contains either of the above, so it's quite likely the variables aren't from the Compiler.

AliasIO:
I had a look into this one but I'm still unsure as well. I'll close this one for now.

PR #1844, opened 2017-11-16
Fix invalid JSON

q--:
apps.json got broken in 8a166daa9


AliasIO:
What an amateur. Thanks!

PR #1845, opened 2017-11-17
Xeora Webserver / Web Framework Addition

freakmaxi:
http://ufal.io
http://jpap.mayadroomajans.com.tr
http://www.mayadroomajans.com.tr/
http://www.resimlerimnette.com/
http://www.mesajkutum.com/
http://www.offlineradio.club/ << it is only used in rest micro service side
http://www.beyinpilim.com/
http://www.jakpark.com/
http://www.bolispirlanta.com/
http://www.xeora.org/


src/apps.json
```diff
@@ -9698,6 +9698,23 @@
...
+			],
+			"html": [
+				"<input type=\"hidden\" name=\"_sys_bind_*\" id=\"_sys_bind_*\" />"
+			],
+			"headers": {
+				"Server": "XeoraEngine",
+				"X-Powered-By": "XeoraCube"
+			},
+			"icon": "xeora.png",
+			"script": "/_bi_sps_v*\\.js",
```
AliasIO: Should this be `"/_bi_sps_v.+\\.js"`?

src/apps.json
```diff
@@ -9698,6 +9698,23 @@
...
 			"website": "http://xenforo.com"
 		},
+		"Xeora": {
+			"cats": [
+				"18",
+				"22",
+				"27"
+			],
+			"html": [
+				"<input type=\"hidden\" name=\"_sys_bind_*\" id=\"_sys_bind_*\" />"
```
AliasIO: Should this be `"<input type=\"hidden\" name=\"_sys_bind_*\" id=\"_sys_bind_[^\"]+\" />"`?

src/apps.json
freakmaxi: actually, the value is like this `/_bi_sps_v1.0.017.js` in the page. I did first like this `/_bi_sps_v(\d\\.\d\\.\d+)\\.js` but I check other entries and changed like that. Should I changed it like yours?

src/apps.json
freakmaxi: and for this, it is the value `<input type="hidden" name="_sys_bind_858321651" id="_sys_bind_858321651" />` in the page. `<input type=\"hidden\" name=\"_sys_bind_(\d+)\" id=\"_sys_bind_(\d+)\" />` was the value and I changed it. I can revert it back to this. what do you say?

src/apps.json
AliasIO: It should be like mine. `v*` makes the `v` optional, `v.+` means `v` followed by one or more characters.

src/apps.json
AliasIO: Either pattern will work, yours is more specific. You don't need the brackets, `(\d+)` is the same as `\d+` in this case.

AliasIO:
Thanks!

PR #1852, opened 2017-11-23
Add Tilda CMS

KarafiziArtur:
Examples of using Tilda CMS:

1. https://socialplus.jp/
2. http://mcr2017.com/
3. http://yeehaacreative.com/
4. https://getcombine.com/
5. http://toktokdoc.com/
6. http://radario.cc/
7. https://chovi.io/
8. http://superheroes.ee/
9. http://7devs.co/
10. http://www.chinesetuxedo.com/


KarafiziArtur:
@gadcam 
> I you escape the dot before the two ws it would be perfect.
Oh, sure. Thanks! Good catch. Done!

AliasIO:
Thanks!

PR #1855, opened 2017-11-24
Escape periods in various regexes

q--:
As you can see in the TravisCI output, I did not escape a few periods because I wasn't sure if they were intentional or not. (And for a couple of patterns, I was sure the period was meant not to be escaped.)


AliasIO:
Nice work, thank you.

PR #1858, opened 2017-11-27
Add styled-components

mxstbr:
[styled-components](https://styled-components.com) is a React.js library for writing CSS in JavaScript. It has almost 11,5k stars on GitHub and is widely known: https://github.com/styled-components/styled-components

A couple of questions:

1. The current pattern which matches the HTML `<style data-styled-components="x">` tags only work on websites that use React server-side rendering. On client-side rendering, those tags don't have the `data-styled-components` attribute. The only way to reliably detect a client-side rendered React app that uses styled-components is that the string `"__styled-components__"` is somewhere in a JavaScript file. I didn't find a way to match the contents of the loaded JS files, so I've left that part out. Is there a way to do that?
2. There is many more libraries like styled-components, which all fall under a bigger category of "CSS-in-JS". (like [aphrodite](https://github.com/khan/aphrodite), [JSS](https://github.com/cssinjs/jss), [Radium](https://github.com/formidablelabs/radium), and [many more](https://github.com/michelebertoli/css-in-js)) Should we add that as a category?


jvoisin:
1. No, I don't think so
2. I don't think it's worth adding a category for this (Also, who the hellish-fuck uses this kind of garbage? No one cares about accessibility or what?!)

mxstbr:
> (Also, who the hellish-fuck uses this kind of garbage? No one cares about accessibility or what?!)

Thanks, I'm sure the dozens of people who've spent years working on this tooling to make developers lives easier (including me) appreciate those words.

While this discussion is way out of scope of this PR, I'd encourage you to read [A unified styling language](https://medium.com/seek-blog/a-unified-styling-language-d0c208de2660) (or [watch the talk](https://www.youtube.com/watch?v=X_uTCnaRe94)) for an understanding of why one would write CSS in JS more generally, and then watch [this talk](https://www.youtube.com/watch?v=MT4D_DioYC8) and [this talk](https://www.youtube.com/watch?time_continue=89&v=bIK2NwoK9xk) to learn more about styled-components specifically.

jvoisin:
It seems nice to be able to write CSS with less hassle, but unfortunately, I'm not really able to use javascript in my default browser, so I simply can't use those websites. Running your system server-side seems like a good idea, but as soon as your trying to run it in my browser, you're making my online life a nightmare :/

Sorry for the harsh word, but from my usage point of view, you're breaking the web.

mxstbr:
@jvoisin I sympathize, but that has nothing to do with `styled-components`, that inherently happens when you use a JavaScript framework and don't server-side render. (in this case React.js but could really be any framework)

AliasIO:
I saw @markdalgleish at Web Directions this year, I thought his talk was the most interesting one by far. I'm a fan.

I don't think the icon works well at 16x16, do you want to replace it with this one perhaps?

styled-components

PR #1898, opened 2017-12-08
Add Mouse Flow detection

fedekau:
This PR adds detection for https://mouseflow.com. You can see their clients in https://mouseflow.com/clients/.


AliasIO:
Thanks!

PR #1899, opened 2017-12-08
Add Inspectlet detection

fedekau:
This PR adds detection for https://www.inspectlet.com/, their clients can be seen there in the main site.


AliasIO:
Thanks!

PR #1900, opened 2017-12-08
Add Amplitude detection

fedekau:
This add detection of https://amplitude.com/ their clients can been in the main website.


AliasIO:
Thank you!

PR #1901, opened 2017-12-08
Add GoSquared detection

fedekau:
This adds detection of https://www.gosquared.com/, their clients can be seen on their site.


src/apps.json
```diff
@@ -11129,6 +11129,18 @@
...
 			"website": "https://akaunting.com"
+		},
+		"GoSquared": {
+			"cats": [
+				"10",
+				"52",
+				"53"
+			],
+			"env": [
+				"_gs;confidence:30"
```
AliasIO: The semicolon needs to be escaped, i.e. `"_gs\\;confidence:30"`. I'll fix this one up.

AliasIO:
Thanks!

PR #1907, opened 2017-12-09
Add new language Portuguese-Portugal

J0N4S13:
This is the translation of the application for portuguese of portugal


AliasIO:
Nice one, thank you!

PR #1911, opened 2017-12-10
Angular version detection

gadcam:
Solves #1434.


gadcam:
@AlexByte Are you sure we need to consider them as different ? Because in https://angularjs.org/ we can read :
>This site and all of its contents are referring to AngularJS (version 1.x), if you are looking for the latest Angular, please visit angular.io.

AlexByte:
I am sure.
> Angular is a complete rewrite from the same team that built AngularJS.

gadcam:
@AlexByte fixed in 27d4552.

AliasIO:
Merged manually, thanks!

AlexByte:
I meant that there should be two JS frameworks in apps.json: AngularJS and Angular.

gadcam:
So the html pattern should point to Angular and the other patterns to AngularJS ?

gadcam:
So the html pattern should point to Angular and the other patterns to AngularJS ?

AlexByte:
Yes

PR #1920, opened 2017-12-13
Fix HTML extraction in WebExtensions

gadcam:
Fix for #1919.
There was a regression, so I implemented again the fix of #615.


AliasIO:
I do still want to cap the size of the HTML that we analyse. Some pages are several megabytes and Wappalyzer will hang.

gadcam:
Then what should be the cap ? 1 MB ? It would be better if we had data on the average page size but I could not find this.

AliasIO:
Very large pages (like Google Sheets or GitHub diffs) often have relatively useless content on them (for us), apart from the very top and bottom of the page. 1MB is huge for a HTML page (1,000,000 chars). The current limit is 500KB, we can make that larger if you think there is a case for it.

gadcam:
The current limit seems to be 50kb. I was only guessing if the 50kb thing was a good choice to optimise the detection according to the Pareto principle, or if another bound would be better. But as I have no data neither on performance nor on html page size I can only guess...

AliasIO:
Thanks, I'll leave the cap as is but use the serializer.

PR #1922, opened 2017-12-13
Update SweetAlert2 to detect .all.js

limonte:
Fix detecting of `sweetalert2.all.js` and `sweetalert2.all.min.js` which were introduced recently.


AliasIO:
Thanks!

PR #1930, opened 2017-12-17
Fixed not working URLs in the webextension options

maxibanki:
Also formatted the code with VS Code, that's why the diff looks so weird.


AliasIO:
Thanks!

PR #1939, opened 2017-12-20
Fixed OpenUI5 Detection (fix #1906)

### LINKED ISSUE #1906: OpenUI5 detection does not work
Hi, 

example page: https://sapui5.hana.ondemand.com/demoapps.html

on which the `sap.ui` javascript variable is set. (check dev tools). See:

https://github.com/AliasIO/Wappalyzer/blob/123058f69b5738dcb156ed145e4c69d7d51cd4d9/src/apps.json#L6189-L6196

But strangely it is not working.

maxibanki:
Fixes the OpenUI5 detection by using the script name instead of checking the javascript variable.

Thanks to @gadcam.
I'm not using the version detection of your regexp because there is not always the version cdn in use. It can also be used a generelly "latest" cdn like this:
https://openui5.hana.ondemand.com/resources/sap-ui-core.js

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -6210,7 +6210,7 @@
 				"12"
 			],
 			"icon": "OpenUI5.png",
-			"env": "sap\\.ui$",
+			"script": "sap-ui-core\\.js",
 			"website": "http://openui5.org/"
 		},
 		"OpenX": {


AliasIO:
This has been merged manually, thank you.

PR #1953, opened 2017-12-22
Webmine detection

kkadosh:
Can be tested here : http://seriesypelis24.com/


AliasIO:
Thanks!

gadcam:
What do you think about scanning LocalStorage ? This one is storing something in a LocalStorage space called webmine.

AliasIO:
Sounds good.

PR #1960, opened 2017-12-26
Enable the use of W in Firefox for Android

gadcam:
This pull request enables to use Wappalyzer in Firefox for Android, before the extension was not showing up because of this bug and I could not spot any error in the console.

However : 
  - it is not possible to change the icon in the URL bar (cf https://bugzilla.mozilla.org/show_bug.cgi?id=1331746)
  - the popup needs a bit of work to be clean in Firefox for Android


AliasIO:
As a Firefox for Android user, very cool. :)

PR #1967, opened 2017-12-28
Change database manager priority

kkadosh:
I think that "database manager" must have higher priority than others categories like "misc", "operating system" ...


AliasIO:
Thanks!

PR #1971, opened 2017-12-30
fix Angular (2+) naming and icon, Angular is not AngularJS

avatsaev:
AngularJS name refers to AngularJS pre 2.0, corrected the icon and naming for Angular 2+ apps


AliasIO:
Thanks!

PR #1987, opened 2018-01-05
Fix: jQuery Migrate detection

maxibanki:
Adding `jQuery.migrateVersion` for version detection via JavaScript: 
https://github.com/jquery/jquery-migrate/blob/180a4531d6efb0c28cbdcef32f0a9033506eac10/README.md#migrate-plugin-api

Example sites on which you can test:
- https://www.tsv-rannungen.de/

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -11312,6 +11312,9 @@
       "cats": [
         "12"
       ],
+      "js": {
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #1988, opened 2018-01-05
added editor schema.json inline support e.g. VS Code, Atom etc.

maxibanki:
With this PR auto completion in the apps.json is possible, when the editor supports it:
image


AliasIO:
Thanks!

Issue #1984, opened 2018-01-05
Change categories priority


kkadosh:
I think that some categories priority have to be changed.
I will update it when i will have some time :)

PR #1990, opened 2018-01-05
Change priorities

kkadosh:
What do you think about those changes ?
Are some priorities too high or too low ?

 #1984


AliasIO:
Looks good!

PR #1995, opened 2018-01-06
Split Angular into AngularJS & Angular

gadcam:
Fix https://github.com/AliasIO/Wappalyzer/pull/1911#issuecomment-354204499
As a sidenote #1971 should not have been merged and should not have passed the CI as some implies field where pointing to AngularJS (but I think it is not checked currently).

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -609,18 +609,25 @@
       "cats": [
         "12"
       ],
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #2001, opened 2018-01-09
Update SweetAlert2 to detect .all.js

limonte:
The same as https://github.com/AliasIO/Wappalyzer/pull/1922

I don't know why the previous change has been undone :confused:


limonte:
Oops, I already submitted the PR which includes this change as well as the change of the plugin's homepage: https://github.com/AliasIO/Wappalyzer/pull/1993/files

PR #2007, opened 2018-01-10
Angular root component can have different selector

sarunint:


AliasIO:
Thanks!

sarunint:
@AliasIO I think you screwed up the merge from master. I've modified the entries of **Angular** but you instead modified the entries of **AngualrJS**, which is two different framework.

#2041 should fix this issue.

sarunint:
I've created a fix at #2042

PR #2012, opened 2018-01-12
Remove pattern with confidence 0

kkadosh:
This pattern has a confidence 0 . I think we can remove it ?


gadcam:
You should not remove it : it is here to provide version detection when we already know that the technology is here.

kkadosh:
Ok thanks for your answer :) i close the PR !

PR #2017, opened 2018-01-14
Add detection for Accelerated Mobile Pages

### LINKED ISSUE #2013: Detect AMP pages
As the title says, it would be great if we could identify [AMP](https://www.ampproject.org/) powered pages.

Example AMP page: https://www.ampproject.org/

igrigorik:
Detect logic is [mirrored from AMP validator](https://github.com/ampproject/amphtml/blob/master/validator/chromeextension/content_script.js#L101).

image

Not sure about the appropriate "cats" values. Is JS frameworks correct, and are there others? 

/cc @pbakaus for a sanity check.

Closes #2013.


pbakaus:
LGTM!

src/apps.json
```diff
@@ -98,6 +98,14 @@
       "icon": "ebis.png",
       "website": "http://www.ebis.ne.jp"
     },
+    "Accelerated Mobile Pages": {
+      "cats": [
+        "12"
+      ],
+      "html": "<html[^>]* amp|⚡",
```
gadcam: I think you need to change it to
```json
"html": "<html[^>]* (?:amp|⚡)",
```
and you could add
```json
"html": "<(?:style )?amp-",
```
Also you could add a detection by js variable
```json
"js": {
     "ampExtendedElements": "",
     "__AMP_BODY_VISIBLE": ""
}
```

igrigorik: > "html": "<html[^>]* (?:amp|⚡)",

Ack, good call.

> "html": "<(?:style )?amp-",

I'm a little hesitant of this one, seems like a honeypot for false-positives. The HTML attribute check is what official AMP validator uses, I think I'd prefer to stay aligned?

> Also you could add a detection by js variable

This seems safe from false-positive side, but do we really need this? If the document doesn't claim to be AMP (per `html` element check above), it's not a valid AMP doc.

AliasIO:
Thanks!

PR #2022, opened 2018-01-16
Escaped a bunch of \'s

brandonkelly:
If these `\`s aren’t escaped property, they will be discarded when the JSON is decoded.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -627,7 +627,7 @@
         "22"
       ],
       "headers": {
-        "Server": "(?:Apache(?:$|/([\\d.]+)|[^/-])|(?:^|\b)HTTPD)\\;version:\\1"
+        "Server": "(?:Apache(?:$|/([\\d.]+)|[^/-])|(?:^|\\b)HTTPD)\\;version:\\1"
       },
       "icon": "Apache.svg",
       "website": "http://apache.org"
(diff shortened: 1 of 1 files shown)


AliasIO:
Right you are. Thank you!

brandonkelly:
@AliasIO looks like a lot of this was undone in 0fbaa9622c0b45efe80630196e07f65c82ff80c9 - was that intentional?

AliasIO:
Not intentional, not sure what happened there. Fixed in fd9c20c700f10796a7aaf429095991c19b25d51c.

brandonkelly:
@AliasIO This was undone again in 0a63ef2ddd3efbf2fee4d3cf6f980792f065dd57. Maybe something is messed up with a build script?

AliasIO:
@brandonkelly Good point, it might be the JSON linter.

brandonkelly:
Will it be fixed soon? Whenever this happens, any app regexes that have `\b` become undetectable by Wappalyzer, as it ends up searching for a litera `b` rather than applying a `\b` (boundary) to the regex. So for example instead of looking for `Craft CMS` with boundaries on each end, it’s looking for `bCraft CMSb`.

brandonkelly:
Thanks!

PR #2023, opened 2018-01-16
Update Craft CMS detection

brandonkelly:
Craft’s `X-Powered-By` header could start including other details besides the product name, so all we should care about is that “Craft CMS” is in there somewhere.


AliasIO:
Thanks!

PR #2027, opened 2018-01-19
Improve Vue.js detection

nickforddesign:
Added watcher for data-v in html elements, now successfully detects Vue.js even if bundled with webpack

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10113,6 +10113,7 @@
       "js": {
         "Vue": ""
       },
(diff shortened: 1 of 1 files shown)


nickforddesign:
Beyond the CI tests, I did test this manually with several different Vue.js projects i work on, some of them webpack based, ann Wappalyzer was able to identify Vue.js correctly in every case.

AliasIO:
Thanks!

PR #2029, opened 2018-01-22
Add Zend version detection

kkadosh:


gadcam:
Can you provide an example website to check if everything is right ? :)

kkadosh:
Yes sorry! Here an example : https://www.aroma-zone.com/ :)

AliasIO:
Thanks!

PR #2033, opened 2018-01-22
Fix problem in version detection via js variables

gadcam:
Fix bug spotted in #2031 :
When the matched element is a number it just returns "True" and not the number.


AliasIO:
Thanks!

PR #2034, opened 2018-01-23
Add an other way to identify redmine

kkadosh:
This can be tested here : https://www.redmineup.com/ and here : https://www.b2brouter.net/fr/


AliasIO:
Thanks!

PR #2041, opened 2018-01-26
Angular and AngularJS is not the same

sarunint:
According to [Angular branding guideline](https://angular.io/presskit#brand-names), AngularJS is referred to 1.x versions and Angular is referred to 2+ versions.

This reverts parts of #2037


AliasIO:
Thanks!

PR #2044, opened 2018-01-28
Add A-Frame detection

gadcam:
A-Frame is Mozilla's VR framework.
You can find showcases here https://aframe.io/blog/


AliasIO:
Thanks!

PR #2045, opened 2018-01-28
AngularJS icon and Angular icon is not the same

sarunint:
Reference: https://angular.io/presskit#examples


gadcam:
I removed it a few weeks ago because I thought in 64x64 we will not notice the difference but I was wrong apparently.

AliasIO:
Thanks!

PR #2096, opened 2018-02-14
fix analyzeMeta duplicate line

totpero:
in analyzeMeta function the r const is declared but never used and in next if is the same code.


AliasIO:
Thanks!

PR #2106, opened 2018-02-16
Add Vaadin

AlexByte:
This can be tested [here](https://ilok.com) or [here](https://www.pepeytono.com)

The Framework has [1300 stars](https://github.com/vaadin/framework)


gadcam:
You could maybe also ad a js pattern <code>vaadin</code>

AliasIO:
Thanks!

PR #2122, opened 2018-02-20
Using `reduce` to remove extra iterations

toastal:
This code iterates a list of links 5 times -- 1 to `from`, 3 to `filter`, and then 1 to `map`. This can be simplified into one `reduce` which iterates the list once.


AliasIO:
Thank you!

PR #2123, opened 2018-02-20
Add guidelines when opening an issue

gadcam:
Helps users to format their bug reports/features request/...

Based on an idea of @Yoginth.


AliasIO:
Cool, thanks.

PR #2129, opened 2018-02-26
Analyze headers

castilloandres:
**What is the issue?**
`analyzeHeaders()` returns a Promise with no value

**What does the PR fix?**
it awaits properly the Promise returned in `analyzeHeaders()` function

**Tests?**
I included some tests for analyzing `html`, `headers`, `js` and `scripts`. It is hard to notice problems in the code without tests since its the case with this `analyzeHeaders` issue.


AliasIO:
Excellent, thanks!

PR #2143, opened 2018-03-06
Fix performance issue

### LINKED ISSUE #2140: Firefox hang when open `Log In` on www.cisco.com
When i open http://www.cisco.com and i then press `Log In` my `Firefox 58.0.2 (64-bit)` will totally stop responding. I can change tabs but if i try to browse it will only start showing that busy line in the tab and nothing more will happening. I can't open any extensions either by clicking on their icons.

image

If i disable Wappalyser 5.4.8 it will work as it should.

gadcam:
Fix #2140, see the issue to have a detailed explanation.
- Remove dangerous scripts patterns
- Avoid to match against plain JavaScript in the js field

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -1520,7 +1520,7 @@
       },
       "html": "<[a-z]+ [^>]*id=\"carbonads-container\"",
       "icon": "Carbon Ads.png",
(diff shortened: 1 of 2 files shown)


AliasIO:
Thanks!

PR #2147, opened 2018-03-07
bootstrap: added version detection

mxschmitt:
This PR adds the detection for the bootstrap.js javascript framework for bootstrap.

Source: https://getbootstrap.com/docs/3.3/javascript/#js-version-nums

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -9769,7 +9769,8 @@
         "<div[^>]+class=\"[^\"]*glyphicon glyphicon-"
       ],
       "js": {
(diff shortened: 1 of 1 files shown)


AliasIO:
Wouldn't this return the version number of the plugin and not Bootstrap itself?

mxschmitt:
Hm, that's true. But in general the bootstrap plugin version should be identical with the version of the bootstrap css framework.
So I think that doesn't make problems, isn't it?

AliasIO:
Assuming that's the case, all good.

PR #2168, opened 2018-03-13
SoundManager: added version detection

mxschmitt:
Adds the version detection for the Soundmanager.
Example URL: a slack workspace
Example version string to verify that the regexp works: `V2.97a.20150601 (HTML5-only mode)`

And yep, the version stored in `soundManager.version` not in the camel case one.


AliasIO:
Thanks!

PR #2172, opened 2018-03-14
Removed % characters in confidence number.

SlivTaMere:
Those % characters are unnecessary and could break compatibility.


AliasIO:
Thanks!

PR #2173, opened 2018-03-14
Fingerprint.js: added version detection

mxschmitt:
Example site where you can test: [superkorb.com](superkorb.com)


AliasIO:
Thanks!

PR #2174, opened 2018-03-15
Update ZURB Foundation icon

colin-marshall:
There is a newer icon for ZURB Foundation since v6 which was been out for a couple years now.


AliasIO:
Thanks!

PR #2175, opened 2018-03-15
detect headers on all content types

miklcct:
Fix for https://github.com/AliasIO/Wappalyzer/issues/1395


AliasIO:
I want to avoid running Wappalyzer over documents like PDFs and other non-HTML content. We can white-list more content types though, including XHTML.

PR #2193, opened 2018-03-21
Snap.svg: added version detection

mxschmitt:
Example site on which you can test it: https://www.wappalyzer.com/technologies/webpack


AliasIO:
Thanks!

PR #2216, opened 2018-04-07
Increase YUI version detection.

AlexByte:


src/apps.json
```diff
@@ -10806,7 +10806,7 @@
         12
       ],
       "js": {
-        "YAHOO": "",
+        "YAHOO.VERSION": "(.*)\\;version:\\1,
```
gadcam: I think you forgot a " at the end of the pattern ;)

AlexByte: Yes

PR #2224, opened 2018-04-10
remove GitPHP

nacimgoura:
I propose we remove that detection. The site is for sale and I did not find the project open source so impossible to have 1000 stars.


gadcam:
👍

PR #2231, opened 2018-04-12
Fix bad category for medium.com

nacimgoura:
Medium is a [blog](https://en.wikipedia.org/wiki/Medium_(website)), but is in the CMS category with Wappalyzer. I corrected that.


AliasIO:
Thanks!

PR #2239, opened 2018-04-16
add backtory

sh-soltanpour:
Backtory is a CDN service provider and provides CDN to web services such as : 
[Beeptunes](https://beeptunes.com)
[Telewebion](http://telewebion.com)
[Torob](https://torob.com)
[Tapsell](https://tapsell.ir)


AliasIO:
Thanks!

PR #2243, opened 2018-04-17
Statcounter detection was missing cases where script tag was output via document.write

eoghanmurray:
E.g. tag code as follows:

```
var sc_project=12847410; 
var sc_invisible=1; 
var sc_security="cad283e9"; 
var sc_https=1; 
var scJsHost = (("https:" == document.location.protocol) ?
"https://secure." : "http://www.");
document.write("<sc"+"ript type='text/javascript' src='" +
scJsHost+
"statcounter.com/counter/counter.js'></"+"script>");
</script>

```

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -9327,6 +9327,11 @@
         10
       ],
       "icon": "StatCounter.png",
+      "js": {
+        "_statcounter": "\\;confidence:100",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks. The 100% confidence tag isn't technically needed as it's the default.

PR #2249, opened 2018-04-18
Change Twemoji cat from JS Graphics to misc

### LINKED ISSUE #2241: Twitter Emoji should not be in the JavaScript Graphics Category
I was checking the most used Graphic Technologies and I have realized that Twitter Emoji is the most used. But I think that library shouldn´t be there. I would only keep the plotting libraries in that category such as D3 and HighCharts. 

image

gadcam:
Fix #2241

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10001,7 +10001,7 @@
     },
     "Twitter Emoji (Twemoji)": {
       "cats": [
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #2277, opened 2018-05-02
fix meta detection

kevNii:
The content variable being in the function-level-namespace led to it being changed before the  asyncForEach with pattern.regex.test(content[2]) was called, leading the wrong results.

### DIFF
--- a/src/wappalyzer.js
+++ b/src/wappalyzer.js
@@ -470,7 +470,6 @@ class Wappalyzer {
     const patterns = this.parsePatterns(app.props.meta);
     const promises = [];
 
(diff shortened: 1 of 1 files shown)


gadcam:
Can you explain the bug more in details ?
Is the problem a race ? If so I think this bug fix is not solving it and we would maybe need to pass <code>content</code> as an argument of the anonymous function in <code>asyncForEach</code>.

AliasIO:
It looks right, each iteration now has it's own instance of `content` instead of re-assigning a value. Nice catch.

PR #2279, opened 2018-05-03
Simplify asyncForEach

gadcam:
I see no obvious reason to do something more complex so I offer that we simplify it like this

### DIFF
--- a/src/wappalyzer.js
+++ b/src/wappalyzer.js
@@ -38,7 +38,7 @@ class Wappalyzer {
   }
 
   asyncForEach(iterable, iterator) {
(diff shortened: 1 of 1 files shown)


AliasIO:
The `setTimeout()` call is what makes it asynchronous. Also the iterators don't return promises so the `Promise.all()` call would be redundant.

gadcam:
I do not get the point.
A Promise is an asynchronous call. What is the benefit to add a 1 ms delay to all Promises ?

AliasIO:
Promises are _not_ asynchronous.

PR #2283, opened 2018-05-04
Remove some legacy code for env detection

gadcam:
- Remove old analyzeEnv function
- Remove useless code in bookmarklet driver


AliasIO:
Thanks!

PR #2290, opened 2018-05-12
fix npm driver boolean CLI arguments

avasilkov:
fix npm driver boolean arguments - Boolean("false") or Boolean("0") is still true, arguments were always true if they were included in the cli independent of values
to test
node index.js http://github.com --debug=0 - debug output
node index.js http://github.com --debug=1 - debug output
node index.js http://github.com - no debug output
same with recursive.


AliasIO:
Thank you!

Issue #2325, opened 2018-05-25
Changed extension options are not applied after closing options tab


antonshevyrin:
**Do you want to request a *feature*, a *new application detection* or report a *bug*?**
Bug
**Is your issue about WebExtension driver (Chrome & Firefox), the website, the NPM driver or the bookmarklet ?**
WebExtension
**What is the current behavior ?**
When I open Wappalyzer options, then check/uncheck "Use technology icon instead of Wappalyzer logo" checkbox and close tab - changes are not applied, even for new tabs.
**If the current behavior is a bug, please provide the steps to reproduce and if possible a demo of the problem.**
1. Right click extension icon
2. Click "Options"
3. Check or uncheck "Use technology icon instead of Wappalyzer logo" checkbox
4. Close options tab. Optionally close options popup before closing tab.
5. Open any webpage (github for example)
6. Observe technology icon option change is not applied

**What is the expected behavior ?**
Option is applied. After closing options tab all new browser tabs must show technology icon in place of Wappalyzer logo or vice versa (depending what was selected in options).
**Which versions of Wappalyzer, and which browser / OS are affected by this issue ? Did this work in previous versions ?**
Wappalyzer 5.4.14
Chrome Version 66.0.3359.139 (Official Build) (64-bit)
Mac OS High Sierra 10.13.4

gadcam:
You are right #2327 should fix it.

gadcam:
You are right #2327 should fix it.
Thank you for your detailed bug report !

AliasIO:
Merged, thank both!

PR #2327, opened 2018-05-28
Fix options page

gadcam:
Bug report: #2325
The value was already correct in local storage but it was not change in <code>options</code> dictionnary in the driver.


AliasIO:
Thanks!

PR #2329, opened 2018-05-29
Add "wappalyzer" command (bin)

martinmaillard:
Wappalyzer can be executed as "wappalyzer <url>" instead of "node index.js <url>".


AliasIO:
Cool!

PR #2332, opened 2018-05-29
Corrected plurals in pt_BR

fernandosavio:
I've made simple adjustments to make all categories follow the same name patterns.
Basically:

- **plurals** to **single**
- **something manager** to **something management**


AliasIO:
Thanks!

PR #2252, opened 2018-04-20
remove regex for YII Framework

yanis-git:
Regex for script detection on YII Framework section prupose to much false positive, as prupose on https://github.com/AliasIO/Wappalyzer/pull/2246 is better to just delete it.

PR #2338, opened 2018-06-11
Update script detection for Yii framework

craiglondon:
This PR #2252 removed the script detection because it conflicted with `https://static.xx.fbcdn.net/rsrc.php/v3/yD/r/vt21ayiiNKR.js`

I have updated the detection so it won't match that but will match the following

```
<script src="/assets/74f24751/yii.js"></script>
<script src="/assets/74f24751/yii.validation.js"></script>
<script src="/assets/74f24751/yii.activeForm.js"></script>
```

Edits and recommendations to the Regex are welcome.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -8991,7 +8991,7 @@
 				"PHP"
 			],
 			"script": "\/yii.*\\.js",
-			"website": "http://yiiframework.com"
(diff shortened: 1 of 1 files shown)


craiglondon:
I will re-create the PR.

PR #2352, opened 2018-06-21
Write "Universal Analytics" in full

q--:
Because you can't really find out what "UA" means if you google it.


AliasIO:
Cheers. :)

PR #2357, opened 2018-06-26
Fix confidenceTotal

PascalHelbig:
The confidence of a pattern is parsed as a string.

For example:
If two matching patterns have the confidence of 10, the confidenceTotal will be calculated by `'10' + '10'`, which will return `'1010'` and Wappalyzer will show it as 100% confidence and not the expected 20%.


AliasIO:
Thanks!

PR #2378, opened 2018-07-09
Update Medium Icon

rizalibnu:
Medium has already replaced the logo in 2017, needed to update in Wapplyzer because still use old logo.

Reference:
https://medium.design/logos-and-brand-guidelines-f1a01a733592


AliasIO:
Thanks!

PR #2387, opened 2018-07-16
Update ByInti Recognition

bloodf:
The last code wasn`t able to recognize the websites running ByInti platform.


bloodf:
The error is on Travis 
image

PR #2404, opened 2018-07-27
Update Ant Design detection regex

yesmeck:
- fix not match `<button type="button" class="ant-btn"`.
- add `drawer`


AliasIO:
Thanks!

PR #2414, opened 2018-08-05
Changing Bubble icon

estraschnov:


AliasIO:
Do you have an SVG version available? If not, would you mind resizing the PNG to 64x64?

PR #2429, opened 2018-08-23
ParcelJS detection added

semoal:
ParcelJS is a webpack alternative.
I didn't find any website on production with parceljs yet, so I created one.
https://min-react-boilerplate-jobdsxjwzf.now.sh/


AliasIO:
Thanks!

PR #2432, opened 2018-08-26
Fix? js confidence calculation

avasilkov:
I'm not sure if this is an error or not, sorry if it isn't.
But js confidence calculation was different from cookie and headers and meta by not providing a key value, even though it has the same schema of a dict in apps.json as headers/cookies/meta
Maybe I'm not seeing something and you can't use string variable name here.


AliasIO:
Thank you!

PR #2437, opened 2018-08-30
Netlify detection improved

semoal:
Improved netlify detection #2434
All the netlify websites contains x-nf-request-id param on headers


AliasIO:
Thanks!

PR #2438, opened 2018-08-30
Fix Issue 2419

kingthorin:
Removed seemingly unnecessary capturing group and non-numeric version tag in Google Analytics pattern to prevent falsely reporting "analytics" or "Universal Analytics" as a version identifier.

I understand that there are `classic` (deprecated) and `universal` flavors offered (at some point) by Google so if you want this split into two different patterns let me know and I can tackle that too. However, with this not being what the majority in the IT industry would classify as a "version" identifier I figured this was a safe way to go.

Fixes AliasIO/Wappalyzer#2419

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -3665,7 +3665,7 @@
         "GoogleAnalyticsObject": "",
         "gaGlobal": ""
       },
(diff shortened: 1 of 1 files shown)


kingthorin:
Hey @AliasIO what do I need to do to get this merged?

AliasIO:
Thanks!

PR #2443, opened 2018-09-06
Update iframe.js

udanpe:
Wappalyzer not work with Firebase
Chrome 68
MacOs 10.13
Uncaught TypeError: Cannot read property 'postMessageId' of null at onPostMessage (iframe.js:1000)


AliasIO:
Thanks!

PR #2447, opened 2018-09-10
[pt-BR] Translated new category entries.

fernandosavio:
Translated "categoryName58" and "categoryName59" new category entries.


AliasIO:
Thanks!

PR #2449, opened 2018-09-11
Delete unexpected coma

kester36:
Fix bug with unexpected coma after the last function parameter
Google Chrome console error:
``` iframe.js:143 Uncaught SyntaxError: Unexpected token ) ```


kester36:
Issue: https://github.com/AliasIO/Wappalyzer/issues/2448

AliasIO:
Thanks!

PR #2467, opened 2018-09-28
[bug fix] Allow analyzeCookies to check for cookies with Uppercase names

brenhom:
Right now analyzeCookies can't handle cookies with uppercase cookie names, such as `AFRAME.version` for A-Frame (since it looks for the lowercase name in `patterns` on line 634). This is a quick fix for that.


AliasIO:
Thanks!

PR #2483, opened 2018-10-06
add tailwindcss

nacimgoura:
I added tailwindcss to the detection list. Tested in Chrome.
github : https://github.com/tailwindcss/tailwindcss


AliasIO:
Thanks!

PR #2486, opened 2018-10-09
improve detection docusaurus

nacimgoura:
This technology is not detected on all sites that use it. 
For example, it is detected on this site (https://draftjs.org/) but not on this one (https://docusaurus.io)

With this improvement, it is detected at more sites.


AliasIO:
Thanks!

Issue #2530, opened 2018-10-31
Wappalyzer not working for bbc.com, wikipedia.org, airbnb.com, ebay.com (and more)


tracykennedy:
While using the NPM driver (wappalyzer version 5.5.4) I have been seeing a lot of timeouts. I ran into these while casually testing my script and it got me worried because I saw it happen so frequently.
My options are set to this:
    ```
const options = {
      debug: false,
      delay: 500,
      maxWait: 60000,
      userAgent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36',
      recursive: false
    };

```
Here are some websites I **can not** seem to get wappalyzer to run successfully on:

- bbc.com
- wikipedia.org
- airbnb.com
- ebay.com

leonelsr:
And it's worth mentioning that it ain't detecting React on https://web.whatsapp.com/ and fully not working on facebook.com

AliasIO:
This should be fixed in the latest release.

leonelsr:
> This should be fixed in the latest release.

Still not detecting React on https://web.whatsapp.com/ and fully not working on https://facebook.com

PR #2539, opened 2018-11-06
Remove split / concat when htmlMaxRows and htmlMaxCols is 0

realhidden:
The htmlMaxCols and htmlMaxRows related regex can take a lot of time, this PR implements a shortcut to ignore that functionality when htmlMaxRows and htmlMaxCols is set to 0. 

This one also solves the problem with #2530 as far as I can tell.


AliasIO:
Thanks!

PR #2560, opened 2018-11-19
Optional use of puppeteer instead of zombie

realhidden:
As an experiment I tried to add the option to replace zombie (which is slow on some sites) with puppeteer.

To use puppeteer, an extra option needs to be added to the Driver creation `{browser: 'puppeteer'}`.

I wanted to keep most of the original code, so created a wrapper for puppeteer, mimicking most of the functions from zombie. Currently 2 options are not handled by the puppeteer code:
- proxy
- username / password

The driver code needed some minor modification to allow browser.visit cb to return a promise, since we need close puppeteer once the analysis is done. Js evaluation is done async (jsAsync), since puppeteer is running in a completely different environment.

Let me know what you think about this approach, since all of this is optional (zombie is still the default browser driver) I hope it can be merged.

This PR also include #2539.


AliasIO:
I like the idea of having an option to change the browser but I think it should be implemented the same way as the [drivers](https://github.com/AliasIO/Wappalyzer/tree/master/src/drivers). We could make `driver.js` more generic and add a `browsers` folder to include browser specific code.

AliasIO:
@realhidden I added the ability to use a different headless browser as I described in my previous comment. This is the wrapper for Zombie.js, on which you can base yours:

https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/npm/browsers/zombie.js

PR #2565, opened 2018-11-23
Less meta regex

realhidden:
Currently we run the same regex on the html multiple times to extract meta fields. Let's extract the metas once, and run the extra, app specific regexes multiple times


AliasIO:
Thank you!

PR #2586, opened 2018-12-16
Change MAINTAINER to LABEL instruction

edsoncelio:
`MAINTAINER` instructions is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated), should use `LABEL` instead `MAINTAINER`.


AliasIO:
Thanks!

PR #2593, opened 2019-01-02
Added Weglot to apps

remyberda:
Weglot is translation API that turn any websites into a multilingual website.
Available on WordPress : https://wordpress.org/plugins/weglot/
Shopify : https://apps.shopify.com/weglot
Or any other technology via Weglot Connect

Weglot is used by 30,000 websites (See active installs https://wordpress.org/plugins/weglot/)


AliasIO:
Thanks!

PR #2599, opened 2019-01-17
Update PHP-Fusion

RobiNN1:


src/apps.json
```diff
@@ -7337,10 +7337,19 @@
       "cats": [
         1
       ],
-      "html": "Powered by <a href=\"[^>]+php-fusion",
+      "html": [
+        "Powered by <a href=\"[^>]+php-fusion",
+        "Powered by PHP-Fusion"
```
AliasIO: HTML patterns should have at least an opening tag (`<`) to avoid it matching plain text.

AliasIO:
Thanks!

PR #2608, opened 2019-01-30
Update platformOS

pavelloz:
Company went through rebranding, hence new icon and one letter changed in http header.


AliasIO:
Thanks!

PR #2622, opened 2019-02-14
Fix script URL technology detection

castilloandres:
**What is the issue?**
`Google Analytics` script URL not detected due to double escaped regex. It was introduced in [this commit](https://github.com/AliasIO/Wappalyzer/commit/eca77528d6c5047d9f1e0b0dc0726189319a59d2#diff-d7f544c6575fa5815df4b466a9e04f40R359).
```
> const doubleEscapedRegex = new RegExp(/google-analytics\.com\\\/(?:ga|urchin|analytics)\.js/i, 'i');
> doubleEscapedRegex.test('http://www.google-analytics.com/analytics.js');
> false

> const singleEscapedRegex = new RegExp(/google-analytics\.com\/(?:ga|urchin|analytics)\.js/i, 'i');
> singleEscapedRegex.test('http://www.google-analytics.com/analytics.js');
> true
```

**What does the PR fix?**
Added single escape instead of double escaped regex so it can detect `Google Analytics` in script URL's.

**Tests?**
I re-added tests for analyzing `html`, `headers`, `js` and `scripts` with specific examples.


castilloandres:
```
/opt/wappalyzer/bin/validate-icons:56
throw new Error('Stray file (no associated app): ' + basePath + file);
^
Error: Stray file (no associated app): src/icons/UserEngage.png
```
@AliasIO Getting this error in Travis CI not sure if I should delete `UserEngage.png` in the icons folder? Looks like its a stray file?

AliasIO:
Thanks!

PR #2628, opened 2019-02-21
Remove stray UserEngage.png file

gnowland:
UserEngage.png was supplanted by user.com.svg in `apps.json` in commit 9cc9e89 and the stray UserEngage.png image file is causing TravisCI and `./run validate` to fail!


AliasIO:
Thanks!

PR #2658, opened 2019-04-10
Add Pantheon detection

kyletaylored:
Pantheon is one of the primary managed hosting platforms for Drupal and WordPress hosting websites for all types of sectors and industries.

- https://pantheon.io/customers (Main site + case studies)
- https://github.com/pantheon-systems
- https://www.datastax.com
- https://www.aclu.org
- https://sendgrid.com
- https://zoo.sandiegozoo.org
- https://www.tableau.com
- https://www.unlv.edu
- https://www.ufc.com
- https://www.opry.com

Pantheon uses their own blend of Varnish they call Styx, which is the primary header to detect.


AliasIO:
Thanks!

PR #2659, opened 2019-04-10
Add Azure detection

kyletaylored:
It probably doesn't need an introduction, but Azure is Microsoft's cloud compute platform similar to AWS or GCP. While working on a couple client projects, I stumbled upon some Azure specific fingerprints that would be worth adding.

- https://azure.microsoft.com
- https://github.com/Azure
- https://github.com/Azure/azure-powershell
- https://www.coyote.com (example for CDN)

**How can we detect**
- For Azure specifically, some named headers are exposed
- For Azure CDN (by Verizon), there are specific headers from their documentation (below)

https://docs.microsoft.com/en-us/azure/cdn/cdn-verizon-http-headers


src/apps.json
```diff
@@ -115,6 +115,30 @@
...
+      },
+      "icon": "azure.svg",
+      "website": "https://azure.microsoft.com"
+    },
+    "Azure CDN": {
+      "cats": [
+        31
+      ],
+      "headers": {
+        "server": "(?:ECAcc|ECS|ECD)",
```
jvoisin: What about `"server": "^(?:ECAcc|ECS|ECD)",` instead, to reduce false-positives?

kyletaylored: Yep, I'm down with that. Will update and run a quick test.

kyletaylored: Azure CDN test

AliasIO:
Thanks!

PR #2667, opened 2019-04-16
Add Tessitura / TNEW

kyletaylored:
Tessitura is a widely used CRM and ticketing management platform for museums, performing arts, zoos, and other cultural and art based institutions around the world. TN Express Web (TNEW) is their companion CMS solution that utilizes the Tessitura API for backend management.

- https://www.tessituranetwork.com
- https://tickets.bclm.com
- https://www.thesmithcenter.com
- https://www.mountvernon.org
- https://www.coloradoballet.org
- https://my.hampsteadtheatre.com
- https://www.alleytheatre.org

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10106,7 +10106,7 @@
       "cats": [
         53
       ],
-      "html": "<*(?:Tessitura Version:) (\\d*\\.\\d*\\.\\d*)?\\;version:\\1",
+      "html": "<!--[^>]+Tessitura Version: (\\d*\\.\\d*\\.\\d*)?\\;version:\\1",
       "implies": [
         "Microsoft ASP.NET",
         "IIS",
--- a/src/apps.json
+++ b/src/apps.json
@@ -10089,6 +10089,35 @@
       },
       "website": "http://telescopeapp.org"
     },
+    "TN Express Web": {
+      "cats": [
+        1
+      ],
+      "headers": {
+        "Set-Cookie": "^(?:TNEW=)"
+      },
+      "cookies": {
+        "TNEW": ""
+      },
+      "implies": [
+        "Tessitura"
+      ],
+      "icon": "tessitura.svg",
+      "website": "https://www.tessituranetwork.com"
+    },
+    "Tessitura": {
+      "cats": [
+        53
+      ],
+      "html": "<*(?:Tessitura Version:) (\\d*\\.\\d*\\.\\d*)?\\;version:\\1",
+      "implies": [
+        "Microsoft ASP.NET",
+        "IIS",
+        "Windows Server"
+      ],
+      "icon": "tessitura.svg",
+      "website": "https://www.tessituranetwork.com"
+    },
     "Tengine": {
       "cats": [
         22
--- a/src/apps.json
+++ b/src/apps.json
@@ -10093,9 +10093,6 @@
       "cats": [
         1
       ],
-      "headers": {
-        "Set-Cookie": "^(?:TNEW=)"
-      },
       "cookies": {
         "TNEW": ""
       },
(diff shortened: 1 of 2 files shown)


kyletaylored:
The main issue I'm having with this one is the cookie detection for TNEW. From all other references I can see, this should be correct, but I can't get it to come up for the life of me. I can see Tessitura working as it should though.

**TNEW cookie being picked up, but not detected**
tnew-cookie

**Tessitura working**
Screen Shot 2019-04-16 at 6 36 22 PM

src/apps.json
```diff
@@ -10089,6 +10089,35 @@
       },
       "website": "http://telescopeapp.org"
     },
+    "TN Express Web": {
+      "cats": [
+        1
+      ],
+      "headers": {
+        "Set-Cookie": "^(?:TNEW=)"
```
jvoisin: Please remove this line, since it's duplicate with the `cookies` one.

src/apps.json
```diff
@@ -10089,6 +10089,35 @@
...
+        "Tessitura"
+      ],
+      "icon": "tessitura.svg",
+      "website": "https://www.tessituranetwork.com"
+    },
+    "Tessitura": {
+      "cats": [
+        53
+      ],
+      "html": "<*(?:Tessitura Version:) (\\d*\\.\\d*\\.\\d*)?\\;version:\\1",
```
jvoisin: This is too false-positive prone, and also disastrous performance-wise, please improve it.

jvoisin:
A brief look at shodan and censys doesn't yield a much results. Do you have an estimation of the number of installation of this CMS worldwide?

kyletaylored:
Thanks for your help! In the culture/art/museum industries, there are generally only a handful of platforms that support membership, ticketing, and donations on a large scale. Some of these include Tessitura, Raiser's Edge (Blackbaud), and PatronManager (Salesforce). I've added another list below, along with the Alexa ranking and some reviews from G2. I couldn't guess how many there are worldwide, other than it's one of the most popular tools in those specific industries.

**Reviews**
- https://www.alexa.com/siteinfo/tessituranetwork.com
- https://www.g2.com/products/tessitura-software/reviews

**More site examples**
- bard.edu
- my.belvoir.com.au
- broadwaytheatrecenter.com
- bsomusic.org
- calhist.org
- calperformances.org
- cerritoscenter.com
- tickets.kcstarlight.com
- nehgs.org
- netsvictoria.org
- newenglandancestors.org
- tickets.newvictory.org
- nycopera.com
- my.oldvictheatre.com
- operasj.org
- orchestraiowa.org
- tessituranetwork.com
- tickets.thegrandwilmington.org
- theparamount.net
- theshed.org
- secure.kimbellart.org

src/apps.json
kyletaylored: Will do. This was my attempt to try and detect either one since `cookies` seemed to fail.

src/apps.json
kyletaylored: Updated this to look specifically for HTML comment blocks to parse (which is the only place this information resides)

kyletaylored:
I figured out why `cookies` doesn't work here. In most Tessitura scenarios, there is the marketing website (WordPress, Drupal, etc), and the ticketing website (TNEW). You can enable session sharing between the two, so while browsing events on the marketing site, a `TNEW` cookie is set, so when you get to the ticketing site, you can have your items already available in your cart.

The limitation in Wappalyzer right now is that it only pulls the cookies for the current domain, but doesn't differentiate between subdomains or root domains. In this case, it looks something like:

1. A cookie is set on `example.com`
2. User goes to `tickets.example.com` to check out
3. Wappalyzer checks `tickets.example.com` cookies, not `example.com` cookies

jvoisin:
Yes, this is how cookies are supposed to work on the web :)

src/apps.json
```diff
@@ -10089,6 +10089,32 @@
...
+        "Tessitura"
+      ],
+      "icon": "tessitura.svg",
+      "website": "https://www.tessituranetwork.com"
+    },
+    "Tessitura": {
+      "cats": [
+        53
+      ],
+      "html": "<!--[^>]+Tessitura Version: (\\d*\\.\\d*\\.\\d*)?\\;version:\\1",
```
jvoisin: Nitpick: something like `"<!--[^>]+Tessitura Version: ([\\d.]+)\\;version:\\1",` would be a bit better :)

kyletaylored:
> Yes, this is how cookies are supposed to work on the web :)

Haha, well yes, except for anyone doing fancy work with CORS. I looked into it a bit last night, and if you tried to match all cookies using a root domain, you run into issues with SLD domains like `.co.uk`. So it's fine, there's no reason to assume Wappalyzer should detect all those. I was able to detect the cookie on another TNEW site, so it's all good!

src/apps.json
kyletaylored: Yea, I was going to do that, but they have these long build numbers at the end that seemed like extraneous info if all that was important was the semver `major.minor.patch` version.

Thoughts?

image

jvoisin: It's ok to show the complete number :)

AliasIO:
Thanks!

PR #2668, opened 2019-04-18
Fix OpenCart false positive

### LINKED ISSUE #2650: google detects as PHP and opencart
**Do you want to request a *feature*, a *new application detection* or report a *bug*?** bug

**Is your issue about WebExtension driver (Chrome & Firefox), the website, the NPM driver or the bookmarklet ?**chrome extension

image

kyletaylored:
This replaces the loose HTML lookup (which just looks for a link with "OpenCart"), which is why it shows up in Google because it's in a search result.

Instead, we can use the OCSESSID cookie that's issued by OpenCart.

**Fixes Google search results**
image

**Works in OpenCart demos**
image

fixes #2650

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -7235,7 +7235,9 @@
       "cats": [
         6
       ],
-      "html": "(?:index\\.php\\?route=[a-z]+/|Powered By <a href=\"[^>]+OpenCart)",
+      "cookies": {
+        "OCSESSID": ""
+      },
       "icon": "OpenCart.png",
       "implies": "PHP",
       "website": "http://www.opencart.com"


AliasIO:
Thanks!

rviscomi:
At [HTTP Archive](httparchive.org) we use Wappalyzer to detect technologies on ~5M websites monthly. We were seeing ~30K websites detected with OpenCart until April, then that number dropped to and stayed at 0.

image

Investigating that brought me to this PR. Is it possible that the `OCSESSID` cookie-based detection is causing too many false negatives?

One issue may be that the cookie is not set on initial page load, perhaps only after some form of user interaction or authentication. All HTTP Archive tests are unauthenticated and home page only. If that's the problem, is there any way to use a different signal that doesn't require auth/interaction?

FWIW looking at the [current sample of OpenCart websites](https://www.wappalyzer.com/technologies/opencart) according to Wappalyzer, many if not all of them seem wrong:

image

kyletaylored:
@rviscomi I think what we can do is in addition to checking for the presence of a cookie, we can also check if the `Set-Cookie` header contains `OCSESSID`, then we don't have to wait for the cookie to be set. Past that, I couldn't find any other distinct detections coming from an OC site that isn't too vague that would likely trigger a false positive.

The list of sites that are listed as a result of the crawl is definitely dirty, and I would almost suggest both Wappalyzer and even HTTP Archive purge or clean the data that was based on the prior detection (it literally only detected a link that contained the OpenCart text).

kyletaylored:
@rviscomi Actually, this doesn't seem possible to access "Set-Cookie" headers through the Browser API.

https://www.w3.org/TR/XMLHttpRequest/#the-getallresponseheaders%28%29-method

> Return all response headers, excluding headers that are a case-insensitive match for Set-Cookie or Set-Cookie2, as a single string, with each header line separated by a U+000D CR U+000A LF pair, excluding the status line, and with each header name and header value separated by a U+003A COLON U+0020 SPACE pair.

Issue #2676, opened 2019-04-19
Missing site cookies from subdomains


kyletaylored:
Unsure if bug, feature, or adjustment.

Currently how Wappalyzer searches for cookies is using the literal hostname of the website, that includes the subdomain. This means if a site utilizes a subdomain (www, docs, etc) and shares a cookie between them, we would never be able to detect that cookie.

An example scenario would be a website utilizing an e-commerce system on a subdomain (like shop.example.com), we wouldn't be able to detect that the website is using that technology.

In an example here, you can see that on this website, there are a number of cookies shared on the domain, but Wappalyzer only picks up a fraction of them that are specific to the domain.

If we go to perotmuseum.org, you can see they use `www` as their primary subdomain, and we use that at the domain key for the cookie search.

**Below: using the literal hostname to search for cookies** 
![Screen Shot 2019-04-19 at 1 18 04 PM](https://user-images.githubusercontent.com/1759794/56439774-fb359300-62ac-11e9-9f5f-52cd9e5dac01.png)

But you can see all the relevant cookies that _could be_ keyed off of by trimming the domain to the root domain.

**Below: all cookies related to the root domain** 
![Screen Shot 2019-04-19 at 1 19 29 PM](https://user-images.githubusercontent.com/1759794/56439810-130d1700-62ad-11e9-914f-69d1781fd7e0.png)

AliasIO:
I think `www` is a legitimate case but a website on a `docs` subdomain is likely to use different technologies than the root domain. Sites on subdomains don't typically set cookies on the root domain.

PR #2677, opened 2019-04-19
Expand cookie domain by parsing for root domain

kyletaylored:
- Implements TLDTS library
- Accounts for SLD domains (.co.uk, .org.nz, etc)
- Returns as part of domain element from parseUrl function

refers to #2676

With this solution, we're able to parse the URL for the root domain, while also accounting for SLD domains. This will give Wappalyzer a broader scope to detect, while also still limiting the information to the target domain.

We're able to achieve this by using the [TLDTS](https://www.npmjs.com/package/tldts) library, which claims to be the fastest domain processing library. 

> `tldts` is the **fastest JavaScript library available for parsing hostnames**. It is able to parse up to 2M hostnames per second on a modern i7-8550U CPU with Node.js version 11.6.0.

This library also has 0 dependencies, making it ideal to reduce overall package size and limit liability.

**Below: showing the spectrum of cookies related to the domain that can be detected**
Screen Shot 2019-04-19 at 1 54 18 PM

### DIFF
--- a/src/drivers/webextension/html/background.html
+++ b/src/drivers/webextension/html/background.html
@@ -5,6 +5,7 @@
 		<meta charset="utf-8">
 
 		<script src="../node_modules/webextension-polyfill/dist/browser-polyfill.js"></script>
+		<script src="../node_modules/tldts/dist/tldts.umd.min.js"></script>
 		<script src="../js/wappalyzer.js"></script>
 		<script src="../js/driver.js"></script>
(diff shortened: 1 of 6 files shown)


AliasIO:
Thanks Kyle! I'm not including this for a couple of reasons. Dependencies complicate the extension review process but I'm also not sure if we should be checking cookies on the root domain. In some cases we might want to but typically a subdomain is considered a different website.

PR #2682, opened 2019-04-24
Add Botble CMS

sangnguyenplus:
Hi,

I want to add Botble CMS (https://botble.com) and rename old company platform cookie's name. 

It will look like this:
Screen Shot 2019-04-24 at 7 09 14 AM

Thanks for a great package!


AliasIO:
Thanks!

PR #2689, opened 2019-05-09
add apigee

Yofial:
apigee is an API Management Platform owned by Google.
The detection is of the API developer portal which is part of the platform and many times public.

Some examples:
- https://developer.cars.com/
- https://developer.accuweather.com/
- https://developer.microbilt.com/
- https://developer.fortellis.io/
- https://dev.telstra.com/


AliasIO:
Thanks!

PR #2692, opened 2019-05-14
Request to add Syncfusion

UmaMaheswariChandrabose:
Syncfusion provides a wide variety of products in all web frameworks that help to develop feature-rich applications - https://www.syncfusion.com/javascript-ui-controls.

Some of the examples are as follows.
https://ej2.syncfusion.com/showcase/typescript/healthtracker/
https://ej2.syncfusion.com/demos/#/material/grid/grid-overview.html 
https://ej2.syncfusion.com/aspnetmvc/DashboardLayout/SEOAnalytics#/material
https://ej2.syncfusion.com/angular/demos/#/material/chart/spline-area
https://ej2.syncfusion.com/vue/demos/#/material/schedule/timeline-resource.html
https://ej2.syncfusion.com/react/demos/#/material/document-editor/default
https://ej2.syncfusion.com/aspnet-core-blazor/Viewer/CustomToolbar
https://ej2.syncfusion.com/aspnetcore/StockChart/HiloOpenClose#/material
https://ej2.syncfusion.com/showcase/angular/stockchart/#/stockChart
https://ej2.syncfusion.com/showcase/vue/assetmanagement/#/
https://aspdotnetcore.syncfusion.com/showcase/aspnetcore/task-planner/project

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10051,7 +10051,7 @@
       "icon": "syncfusion.svg",
       "website": "https://www.syncfusion.com/javascript-ui-controls",
       "js": { "ej.base": "" },
-      "html": "<[^<]* [class=\"[^<]*(\be-control\b) [^<]* (\be-lib\b)[^<]*\" [^<]*>"
+      "html": "<[^<]* class=\"[^<]*\be-control\b[^-][^<]*\be-lib\b[^>]*\"[^>]*>"
     },
     "Synology DiskStation": {
       "cats": [
--- a/src/apps.json
+++ b/src/apps.json
@@ -3121,6 +3121,15 @@
       "icon": "Erlang.png",
       "website": "http://www.erlang.org"
     },
+    "Essential JS 2": {
+      "cats": [
+        12, 18, 59
+      ],
+      "icon": "syncfusion.svg",
+      "website": "https://www.syncfusion.com/javascript-ui-controls",
+      "js": { "ej.base": "" },
+      "html": "<[^<]* class=\"[^<]*\be-control\b[^-][^<]*\be-lib\b[^>]*\"[^>]*>"
+    },
     "Etherpad": {
       "cats": [
         24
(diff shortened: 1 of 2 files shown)


UmaMaheswariChandrabose:
Hey @AliasIO ,

Can you please check and merge this pull request?

Regards,
C. Uma Maheswari

AliasIO:
Thanks!

src/apps.json
```diff
@@ -3121,6 +3121,15 @@
       "icon": "Erlang.png",
       "website": "http://www.erlang.org"
     },
+    "Essential JS 2": {
+      "cats": [
+        12, 18, 59
+      ],
+      "icon": "syncfusion.svg",
+      "website": "https://www.syncfusion.com/javascript-ui-controls",
+      "js": { "ej.base": "" },
```
kyletaylored: I'm almost positive this is wrong. Even on the current Syncfusion site, there is no property of `ej.base`, but there is `ej.version`, which should likely be used.

src/apps.json
```diff
@@ -3121,6 +3121,15 @@
...
+      "html": "<[^<]* class=\"[^<]*\be-control\b[^-][^<]*\be-lib\b[^>]*\"[^>]*>"
```
kyletaylored: I couldn't get this to trigger on their demo site. I don't think this detection was actually validated.

UmaMaheswariChandrabose: Hi @kyletaylored ,

I have validated this expression in regex validator and it detected the class names exactly. What I tried here is to match the HTML elements whose class name contains the term "e-control" and "e-lib". Is there any other way to validate this?

src/apps.json
UmaMaheswariChandrabose: Thanks for your suggestion @kyletaylored , but ej.version is not a part of our brand new Essential JS 2 product, so we can't use it. Anyhow, I'll revisit this JavaScript method name. Thanks again.

kyletaylored: Ah, I see. Then I'd add both to pick up both new and old versions.

src/apps.json
kyletaylored: The best way is to build the extension locally and add it to Chrome, the refresh the extensions and test that way using the background window debugger (for the extension, not the webpage).

Issue #2375, opened 2018-07-04
NPM driver returns inconsistent results


xde013:
## What is the current behavior ?
Wappalyzer fails sometimes to get response from server and thus outputting empty result. 
Other times it returns incomplete list of detected services.

**Empty result**
![image](https://user-images.githubusercontent.com/21088699/42273850-c1e506ea-7f8a-11e8-8eee-9f0ed4981025.png)

**Incomplete result**
![image](https://user-images.githubusercontent.com/21088699/42277856-16cdee70-7f99-11e8-8217-f757779f791d.png)

**the steps to reproduce the problem**

```
const url = 'https://www.engadget.com'
const wappalyzer = new Wappalyzer(url, options); 
wappalyzer.analyze()
  .then(json => {
    process.stdout.write(JSON.stringify(json, null, 2) + '\n') 
    process.exit(0);
  })
  .catch(error => {
    process.stderr.write(error + '\n')
 
    process.exit(1);
});
```
**What is the expected behavior ?**

*Output complete list of all services*

![image](https://user-images.githubusercontent.com/21088699/42277887-309f87aa-7f99-11e8-8e15-bad50140e6a4.png)

AliasIO:
I'm not sure what causes this. It's an issue with Zombie.js most likely.

xde013:
@AliasIO is there any possibility to switch to another browser?

AliasIO:
What do you suggest?

dege88:
Headless chrome is really nice to work with. Especially with puppeteer ( https://github.com/GoogleChrome/puppeteer ).
Unfortunately I've some issues with zombie too (it's crashing the entire node process on some websites)

AliasIO:
Zombie.js is much, much leaner than headless Chrome and runs entirely in JavaScript, making it more portable.

AliasIO:
It's now possible to use a different headless browser by crearting a wrapper, like this one for Zombie.js:

https://github.com/AliasIO/Wappalyzer/blob/master/src/drivers/npm/browsers/zombie.js

juffardm:
Hi, I try to use Wappalyzer with Puppeteer. Do you have a working wrapper for this browser? :)
(I have troubles to get the js of the page I want to analyze)

ThomasLohner:
@juffardm Did you succeed? Can anyone share a working example of using wappalyzer with puppeteer?

ThomasLohner:
[...]

Issue #2421, opened 2018-08-17
Inconsistencies between Chrome extension and npm package


cangoektas:
I found that for some websites the npm package produces different results than the Chrome extension. Specifically, the npm package is unable to detect Google Analytics while the Chrome extension is.

I'm using the `5.5.3` in Chrome and `5.5.1` for the npm package. I also tried different options for the package but nothing really helped.

### Steps to reproduce
1. Visit http://runifico.com/ or http://www.der-autotester.de/
2. Check the results of the Chrome extension (it detects Google Analytics)
3. Run the following JavaScript snippet and check the results:

```js
const options = {
  debug: true,
  delay: 0,
  maxDepth: 0,
  maxUrls: 10,
  maxWait: 5000,
  recursive: true,
  htmlMaxCols: Number.MAX_SAFE_INTEGER,
  htmlMaxRows: Number.MAX_SAFE_INTEGER,
  userAgent: 'Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2228.0 Safari/537.36'
};

const wappalyzer = new Wappalyzer('http://runifico.com/', options);

wappalyzer.analyze().then(console.log);
```

<details>
<summary>The results will not contain Google Analytics</summary>

```json
{
  "urls": {
    "http://runifico.com/": {
      "status": 200
    }
  },
  "applications": [
    {
      "name": "Apache",
      "confidence": "100",
      "version": "2.4.34",
      "icon": "Apache.svg",
      "website": "http://apache.org",
      "categories": [
        {
          "22": "Web Servers"
        }
      ]
    },
    {
      "name": "Font Awesome",
      "confidence": "100",
      "version": null,
      "icon": "Font Awesome.png",
      "website": "http://fontawesome.io",
      "categories": [
        {
          "17": "Font Scripts"
        }
      ]
    },
    {
      "name": "Google AdSense",
      "confidence": "100",
      "version": null,
      "icon": "Google AdSense.svg",
      "website": "https://www.google.fr/adsense/start/",
      "categories": [
        {
          "36": "Advertising Networks"
        }
      ]
    },
    {
      "name": "Google Font API",
      "confidence": "100",
      "version": null,
      "icon": "Google Font API.png",
      "website": "http://google.com/fonts",
      "categories": [
        {
          "17": "Font Scripts"
        }
      ]
    },
    {
      "name": "PHP",
      "confidence": "100",
[...]

PR #2711, opened 2019-06-04
Puppeteer support

realhidden:
There was an old PR (#2560) for adding Puppeteer as a supported browser, using the new Zombie wrapper I updated the code


ThomasLohner:
I'd really like to see this merged. This fixes #2375 and #2421 .

Vasile-Peste:
Updates?

AliasIO:
Thanks for the PR, and sorry for neglecting it for so long. I'd prefer if this was a separate package that can be included as a dependency, rather than merging and supporting it. I'd be happy to link to it from the README.

PR #2716, opened 2019-06-12
Add more cookies to identify VTEX

brunojdo:
VTEX is the biggest e-commerce player in LATAM and we're adding more ways to identify our platform.


AliasIO:
Thanks!

PR #2717, opened 2019-06-12
Update Doxygen homepage

helaan:
Because the previous hoster decided to disband itself, Doxygen moved to its own domain. Site linked on https://github.com/doxygen/doxygen


AliasIO:
Thanks!

PR #2718, opened 2019-06-13
Update detection for Shopify storefronts

t-kelly:
Followup to email from @rviscomi about [Web Almanac](https://github.com/HTTPArchive/almanac.httparchive.org)

Updates Shopify detection with:
- Reduced confidence with file references to `cdn.shopify.com`
- Reduced confidence with declaration of `window.Shopify`
- Look for `x-shopid` and `x-shopify-stage` headers
- If origin is `myshopify.com`, then we know the website is built with Shopify.

This should also remove `shopify.com` websites (used for Shopify marketing) from detection -- does that make sense for Wappalyzer?

I ran `./run validate` but had several errors from other applications that I didn't touch? 🤷‍♂️

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -9274,11 +9274,16 @@
       "cats": [
         6
       ],
-      "html": "<link[^>]+=['\"]//cdn\\.shopify\\.com",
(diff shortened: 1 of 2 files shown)


AliasIO:
Thanks!

PR #2721, opened 2019-06-19
Make AMP detection slightly more robust

pbakaus:
The AMP detection causes false positives with sites that didn't have the amp attribute in the `<html>` tag, but loaded an AMP related script, as AMP's JS adds dynamic amp-* attributes to the `<html>`.

This also updates our name (now just AMP), and the website (now amp.dev).

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -107,13 +107,13 @@
       },
       "website": "http://awstats.sourceforge.net"
     },
-    "Accelerated Mobile Pages": {
+    "AMP": {
       "cats": [
         12
       ],
-      "html": "<html[^>]* (?:amp|⚡)",
+      "html": "<html[^>]* (?:amp|⚡)[^-]",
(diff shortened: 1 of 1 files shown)


src/apps.json
```diff
@@ -107,13 +107,13 @@
       },
       "website": "http://awstats.sourceforge.net"
     },
-    "Accelerated Mobile Pages": {
+    "AMP": {
```
rviscomi: FYI this might break continuity of some HTTP Archive queries.

pbakaus: Mhh, is there anything we can do about that, whilst still changing the name?

rviscomi: Don't let this stop the renaming, we'll have to figure out a way to coalesce both names back to the same technology on our end.

AliasIO:
Thanks!

PR #2722, opened 2019-06-19
Add Derak.Cloud CDN & Cloud Services

AmirBitaraf:
Hi there,
We are a Cloud Solution Provider and CDN called Derak.Cloud. We are currently servicing 50+ websites. This pull request contains our application.
Thanks


AliasIO:
Thanks!

PR #2730, opened 2019-06-30
Update styled-components identification on injected CSS

brohlson:
Noticed styled hasn't been being registered on most of the projects I use it on.   On SSR projects, all of the generated hashed classes will have the following declaration:

Example:
`/* sc-component-id: sc-1v848in-0 */`


AliasIO:
Thanks!

PR #2738, opened 2019-07-09
add npm @sentry/browser

osben:
src https://github.com/getsentry/sentry-javascript/blob/8e08b297d3d28563ad80d9c2a671e59e2f78404f/packages/utils/src/logger.ts#L61


AliasIO:
Thanks!

PR #2762, opened 2019-09-04
Add Sapper, Updated Svelte regex

jhewt:
# Sapper
> Sapper is a framework powered by Svelte for building web applications of all sizes, with a beautiful development experience and flexible filesystem-based routing.

> Unlike single-page apps, Sapper doesn't compromise on SEO, progressive enhancement or the initial load experience — but unlike traditional server-rendered apps, navigation is instantaneous for that app-like feel.

https://sapper.svelte.dev/

# Svelte
The previous Svelte regex detection gives false positives when a page contains a class like `t-svelte-something`. An example of this is StackOverflow, a simple "svelte" search gives a lot of divs with similar class names.

A better way to detect that is to just check the last class on an element, Svelte's compiler always appends the compiled class.

However, it's very hard to detect if a page is built with Svelte since there's no global variables or other than the compiled class names, but those can be absent if the developer uses global css styles.


AliasIO:
Thanks!

PR #2769, opened 2019-09-11
CrossBox fix

lowshadow:
Correction to detect Crossbox via the "server" header.  The current HTML match doesn't work for the white-labeled instances.


AliasIO:
Thanks!

PR #2783, opened 2019-09-24
Add theme support for dark mode

### LINKED ISSUE #2776: Feature Request: Dark Mode
Description
-

**Do you want to request a *feature*, a *new application detection* or report a *bug*?**
Feature

**Is your issue about WebExtension driver (Chrome & Firefox), the website, the NPM driver or the bookmarklet ?**
Web extension

**What is the current behavior ?**
The extension popup has a light theme regardless of the browser setting.

**What is the expected behavior ?**
The extension popup switches to a dark theme when the browser uses one.

**Which versions of Wappalyzer, and which browser / OS are affected by this issue ? Did this work in previous versions ?**
This affects all versions of Wappalyzer running in browser supporting the css `prefers-color-scheme`-media query.

Reasoning
-

Chrome and Firefox now both support the corresponding ability to query a users color preference and allow one to change the color scheme based on this. I would like to implement this functionality into the extensions popup window as many website can be styled with other extensions but extensions themself can't be styled by 3rd party extensions.

I would be willing to work in this issue in my free time if this feature is of interest
-

I can't guarantee any deadline though I'd like to look into the technical details of implementing such a dark mode.

Possible Obstacles
-

As already mentioned in #2757 some icons are poorly visible on a dark background. This though doesn't only applies to a potential dark mode but also the preview in the extension tray. It would then however also extend to the popup area.
There are three possible ways to solve this problem i thought of:
1. Add a white background to the icons in the popup
2. Provide alternative icons for problematic logos - these then need to be chosen dynamically
3. Fix the problematic logos by e.g. adding a corresponding background or outline

Personally I think option 3 would be easy to implement and the best for the long run. This would also resolve issue #2757 in the hotbar.
At first i would try to implement the first option because it would not involve going through all ~1000 icons. If possible we could also try to draw the icon on a white canvas and use that as a toolbar icon to resolve the above issue. This however seems like a hack and fixing the symptoms instead of the cause.

I would like to hear your feedback on this... :D

kyletaylored:
- fixes #2776

### Preview
Dark mode support

### DIFF
--- a/src/drivers/webextension/css/popup.css
+++ b/src/drivers/webextension/css/popup.css
@@ -210,6 +210,9 @@ body {
   .detected__category-link, .detected__app {
     color: white;
   }
+  .detected__category-link:hover {
+    color: #B3B3B3;
+  }
   .detected__app-icon {
     -webkit-filter: drop-shadow(0px 0px 1px rgba(255,255,255,1));
     filter: drop-shadow(0px 0px 1px rgba(255,255,255,1));
--- a/src/drivers/webextension/css/popup.css
+++ b/src/drivers/webextension/css/popup.css
@@ -108,7 +108,6 @@ body {
 	margin-right: .5rem;
 	vertical-align: -.2rem;
 	width: 16px;
-	filter: drop-shadow(5px 5px 5px #222);
 }
 
 .detected__app-name {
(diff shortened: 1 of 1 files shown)


AliasIO:
Nice one, thanks!

PR #2791, opened 2019-10-08
Added and updated a list of things.

Stryxus:
- Resized all oversized icons to 64x64, this will reduce the size of the extension and fix possible rendering issues.
- Added Blazor support, this support only covers the server-side rendering variant.
- Updated most, if not all Microsoft based icons, including Office 2019 icons and Windows Server.
- ASP.NET usage no longer shows IIS usage due to the fact that IIS isnt the only thing which is used now.
- Added CSharp language support.


AliasIO:
Why were FAST and MS Sharepoint removed?

Stryxus:
As far as i know, sharepoint doesnt exist anymore on web. If im wrong then this can be easily corrected.

Runiko:
> 
> 
> As far as i know, sharepoint doesnt exist anymore on web. If im wrong then this can be easily corrected.

I think I saw some pages built on sharepoint.

Stryxus:
If there can be any pages found which still use sharepoint (And arent horribly broken obviously) then ill update all of this and fix the conflicts.

kyletaylored:
> And arent horribly broken obviously

That's not really a validation to remove a detection. I've seen a lot of very broken WordPress sites, but obviously we're not going to remove them. SharePoint is (unfortunately) alive and well, usually implemented in higher education and government sectors. This data would be valuable to consultants or agencies who help companies migrate Sharepoint to other CMS platforms as part of an outbound sales strategy.

https://trends.builtwith.com/websitelist/Microsoft-Sharepoint/Historical

Also the Kestrel server is specific to .NET Core, not just .NET as a framework.

Stryxus:
@kyletaylored .NET Core will be .NET so better to do it now than add confusion.
Plus Wordpress isnt really a comparable thing since the popularity of sharepoint and wordpress is completely different and well... Just telling the tech used in broken, outdated websites seems useless.

But i do see the organisational use of it. When i can, ill re-add it. No rush since Wappalyzer doesnt update that frequently.

kyletaylored:
I'm going to reject this, but I have a request. Instead of bundling all of these random fixes together into a single PR, put them into individual / logical groups so it's easier to digest.

- Create an issue for the performance pieces (resizing icons) to be addressed.
- Create an issue for updating the Microsoft product references / renames with info.
- Create a separate PR just for adding Blazor / CSharp

Thanks!

PR #2804, opened 2019-10-22
Better Click Funnels Detection

grallc:
Hello,
I noticed that, on some websites, the ClickFunnels detection wasn't working.
Example :
- https://go.iag-media.com/kaizencure

It's because of the fact that their scripts' src are not using clickfunnels domain but their one, therefore the detection of clickfunnels using those scripts isn't working.

Here is another fix that should work on every clickfunnels website.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -1481,8 +1481,8 @@
       "cats": [
         32
       ],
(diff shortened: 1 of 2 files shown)


AliasIO:
Thanks!

PR #2806, opened 2019-10-24
Improve Carbon Ads script pattern

q--:
Carbon Ads now uses *cdn.carbonads.com*, which isn't matched yet. This pattern will match all scripts loaded from carbonads.com and any subdomains. HTTP, HTTPS and protocol-relative URLs will match.

Not overly broad because the informational website is at carbonads**.net**


AliasIO:
Thanks!

PR #2810, opened 2019-10-25
Improve webtrekk detection

q--:
The variable `webtrekk` is only present on a few websites, it seems.
Adding a bunch of other Webtrekk variables to improve detection


AliasIO:
Thanks!

PR #2811, opened 2019-10-25
Fix and improve VideoJS detection

q--:
Current VideoJS versions always set a lowercase `videojs` variable.
Maybe they always have, variable names weren't case sensitive before the
transition to using `"js"` in apps.json. Still, preserving the
mixed-case name as well just in case

Also added a new `script` pattern for detection, and a fresh SVG logo


AliasIO:
Thank you!

PR #2812, opened 2019-10-27
fix example in npm readme.md

petskratt:
capitalization fix -

while class Wappalyzer defines “this.browser = ZombieBrowser;” its constructor returns instance of Driver, that defines “this.Browser = Browser;”


AliasIO:
Thanks!

PR #2818, opened 2019-11-02
wpJet

wpJet:
wpJet Cache & Monitoring Software PaaS https://wpjet.xyz


kyletaylored:
Can you list at least 5 public sites that are using wpJet to validate the usage? I don't see how the service is a PaaS, so more information on that would be useful as well.

wpJet:
website  wpjet.xyz 7 days only online , is a new project
please try free my web software
https://wpjet.xyz/download/

thanks

kyletaylored:
I'm going to reject this PR on a number of grounds, and hopefully you'll understand these reasons:

- Wappalyzer detects thousands of apps (as shown on the [technologies](https://www.wappalyzer.com/technologies) page).
- The business model for Wappalyzer, outside of providing an open source digital fingerprinting service, is to provide reports / data of sites using one of those applications. While customers may want this data for a number of reasons, it's often used for sales research.
- New applications need to be "somewhat" popular in order to be useful. If no one is using the service, then no data will be reported or useful, but Wappalyzer still takes a performance hit / becomes bloated as it processes all the regex matches in the apps file.

While I commend you for creating a new service offering, it doesn't meet a reasonable threshold to include it in Wappalyzer yet. Obviously keep working on it, and in time, come back and re-submit a PR with updated usage stats and examples!

PR #2820, opened 2019-11-05
Add option to enable dark mode compatibility

### LINKED ISSUE #2814: Feature Request: config option to choose between dark and light mode regardless of the browser theme
IMHO currently the dark theme looks very poor, and I'd like to have the ability to revert to light mode without switching the browser theme to a light one.

kyletaylored:
- The dark mode update introduced an undesirable issue where it would force browsers with dark theme enabled to use the new dark mode, even though it needs some work. (#2776)
- This update will add a config option to allow users to enable dark mode compatibility themselves, without forcing dark mode users into using it.
- Removed the dropshadow for icons. Very few icons were dark enough for it to actually be useful, but more often than not it was an eyesore.
- fixes #2814

Nov-04-2019 23-41-03


AliasIO:
Thanks for your work on this Kyle, your branch went missing but I implemented these changes.

PR #2823, opened 2019-11-06
Add MobX

q--:
Note: `__mobxGlobals.version` appears to be the version of a specific MobX
component, as it was 4 at MobX version 2.x and went to 5 at MobX 3.x
(and has stayed 5 in all subsequent versions).

Because of this I haven't used the variable for version detection


kyletaylored:
Seems popular and well used - great add!

PR #2825, opened 2019-11-06
Fix HTML pattern for Vue.js

q--:
A non-capturing group was accidentally not marked as optional

Also make it a little more strict by ensuring there is a ` ` or other whitespace before the attribute


AliasIO:
Thanks!

PR #2827, opened 2019-11-06
Add a new way to detect Lodash

q--:


kyletaylored:
@q-- Could you verify that detection for me?

PR #2842, opened 2019-11-09
Fix Backdrop CMS detected as both Drupal and Backdrop

klonos:
Fixes https://github.com/AliasIO/Wappalyzer/issues/2841

(more may need to be done?)


AliasIO:
Thanks!

PR #2846, opened 2019-11-14
[TASK] Add some additional checks for TYPO3

CDRO:
TYPO3 CMS is not properly recognized when the resources are loaded using absolute URIs.

In order to prevent this, an optional slash has been added to the checks for `link`, `img` and `script` tags.

Also an additional rule to check if the source code contains the famous "This website is powered by TYPO3" comment at the top.

The JavaScript function `decryptString` is used by TYPO3 CMS to decrypt encrypted `mailto:` urls in the page and is available at any time. I'm pretty confident that TYPO3 is the only CMS using this function with this name but I'll gladly be proven wrong if anybody knows something else. If that's the case, we could change the confidence to 50 or less (or remove this check completely).

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10713,14 +10713,15 @@
         1
       ],
       "html": [
-        "<link[^>]+ href=\"typo3(?:conf|temp)/",
-        "<img[^>]+ src=\"typo3(?:conf|temp)/"
+        "<link[^>]+ href=\"/?typo3(?:conf|temp)/",
+        "<img[^>]+ src=\"/?typo3(?:conf|temp)/",
+        "<!--\n\tThis website is powered by TYPO3"
       ],
       "icon": "TYPO3.svg",
       "implies": "PHP",
-      "script": "^typo3(?:conf|temp)/",
+      "script": "^/?typo3(?:conf|temp)/",
       "meta": {
-        "generator": "TYPO3\\s+(?:CMS\\s+)?([\\d.]+)?(?:\\s+CMS)?\\;version:\\1"
+        "generator": "TYPO3\\s+(?:CMS\\s+)?(?:[\\d.]+)?(?:\\s+CMS)?"
       },
       "url": "/typo3/",
       "website": "https://typo3.org/"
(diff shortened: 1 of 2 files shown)


src/apps.json
```diff
@@ -10713,12 +10713,16 @@
...
+        "<link[^>]+ href=\"/?typo3(?:conf|temp)/",
+        "<img[^>]+ src=\"/?typo3(?:conf|temp)/",
+        "<!--\n\tThis website is powered by TYPO3"
       ],
       "icon": "TYPO3.svg",
       "implies": "PHP",
-      "script": "^typo3(?:conf|temp)/",
+      "js": {
+        "decryptString": ""
+      },
```
kyletaylored: I know you said that AFAIK that TYPO3 is the only CMS using this global function, but `decryptString` such a vague name that I wouldn't put it past anyone creating that kind of simple function on a site, it's just too prone to false positives. Everything else looks good, so if you can remove that, I'll merge this in. Thanks!

CDRO: Sure thing, will update asap.

CDRO:
I now removed the check for the `decryptString` function, as suggested by @kyletaylored. I also update the generator detection, since there does not seem to be any version I could find that acutally uses the `;version:xyz` format, I made it optional, I hope everybody can agree, that this is globally more accurate, since TYPO3 does not provide it's version number in the generator meta tag since version 4.5.x

CDRO:
I don't know why the test failed, any ideas?

kyletaylored:
@cdro You have a regex syntax error. Be sure to use the validation tool prior to the PR. You can use it from the root Wappalyzer directory running `./run validate`

The error is this:

```
SyntaxError: Invalid regular expression: //TYPO3\s+(?:CMS\s+)?([\d.]+)?(?:\s+CMS)?(?://: Unterminated group
```

CDRO:
That's very strange... I tested this exact regex with regex 101 and some other tools and it passed easily, I'll pass it through `./run validate` and update my code asap.

kyletaylored:
Perfect - thanks!

PR #2856, opened 2019-11-20
Added Javadoc

q--:
Javadoc doesn't have its own icon, so I went with the Java icon instead.

Or should I use the default icon?


kyletaylored:
It looked like this might be the unofficial logo for Javadoc? I couldn't find a vector, but I did my best recreating one.

[!Javadoc](https://svgshare.com/i/GG4.svg)

q--:
https://www.oracle.com/java/duke.html Had  to look it up, but that's Duke, Java's mascot.

So, not really unique to Javadoc. Probably not seen much anymore because Oracle hasn't done much about promoting it. I could use it instead of the Java icon if you prefer it, though.



[I found an SVG version at Wikipedia as well](https://en.wikipedia.org/wiki/File:Duke_(Java_mascot)_waving.svg), but yours would be fine too.

PR #2860, opened 2019-11-25
Add DatoCMS

matjack1:
I'm suggesting to add DatoCMS to the list of recognised CMSs.

Here's a list of sites using Dato: https://www.datocms.com/use-cases/

The core of DatoCMS is closed source so I cannot reference anything there, but if you need more info please ask. I think in general is worth to add as we are currently counting more than 10k projects started and we are steadily growing with numbers.

Thank you!


AliasIO:
Thanks!

PR #2881, opened 2019-12-06
Detect Ez Platform, update (legacy) Ez Publish URL

q--:
Ez Publish is now a legacy application, Ez Platform is meant to replace it.

Added Ez Platform as a separate application, because they seem technologically different.
(Platform is built around Symfony, Publish is not as far as I can tell but uses a few Symfony components.)


AliasIO:
Thanks!

PR #2882, opened 2019-12-06
Fix Symfony version detection

q--:
Changed	`(?:<other patttern>|<pattern with version detection>)`
To	`(?:<pattern with version detection>|<other pattern>)`


AliasIO:
Thanks!

PR #2883, opened 2019-12-06
Fix WebsiteBaker URL, add version detection

q--:
Also added a few alternative detection methods


q--:
The cookie name doesn't support RegEx, right? Otherwise the following would work:

`wb[\\-_]\\d{4}[\\-_]sid`

AliasIO:
Oops, 20K changes. Mind fixing that up? :)

PR #2892, opened 2019-12-20
Add Strikingly

bobywoodwarrior:
Hi,

I'm suggested to add Strikingly to the list, which is a CMS.

You can find a list of sites using Strikingly here : https://www.strikingly.com/s/discover.

Thank you !


AliasIO:
Thanks, but your PR contains over 5,000 changed lines. I'm guessing it's auto-indented.

PR #2893, opened 2019-12-22
Puppeteer support

realhidden:
The current structure and window prop evaluation magic is clearly not a clean fit for puppeteer. Still, third time (#2711, #2560) I try to be as lean as it gets. 

A simple puppeteer based browser implementation on the new Browser class, with a "window object" approximation using a shallowMap function, one peer dependency and **no modification** in the core.

Let me know if this direction is a fit for the npm part of this project, I'm more than happy to clean up the code, document extra options, etc.

### DIFF
--- a/src/drivers/npm/browsers/puppeteer.js
+++ b/src/drivers/npm/browsers/puppeteer.js
@@ -1,0 +1,199 @@
+const { Cluster } = require('puppeteer-cluster');
+const Browser = require('../browser');
+
+let cluster = null;
+
+function puppeteerJsEvalFunction() {
+  const shallowMap = (origin, level) => {
+    try {
+      if (level === 0) {
+        return true;
+      }
+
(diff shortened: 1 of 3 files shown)


AliasIO:
Thanks for the PR, I appreciate the work you've done on this. I simplified your code in 635f6871d4edc3f53c4d4677841d3adbadcf2d20, let me know what you think. I didn't see the need for puppeteer-cluster as a calls to `visit()` never run more than one instance.

I noticed the results aren't exactly the same as when Zombie.js is used, I believe that's because Puppeteer doesn't wait for all JavaScript to be executed on the page. E.g. Snap.svg is not being picked up on https://www.wappalyzer.com.

I'll close this PR but happy to keep the conversation going here.

AliasIO:
Never mind, fixed it in a4b6cee3fdd46e5ca3f99c5d3e80ee32a960a238. Should be all good to go now.

PR #2894, opened 2019-12-29
Update Materialize implications

Gildas-GH:
As of version 1.0, jQuery is no longer a dependency of Materialize.


AliasIO:
Thanks!

PR #2903, opened 2020-01-03
Added CoreMedia CMS detection to apps(https://www.coremedia.com/)

greenoid:
CoreMedia CMS is a software product of a company based in Hamburg, Germany.
The CMS is now rebranded as CoreMedia Content Cloud (CMCC) since version 10.
Many big companies are using it, see 96 of them:
https://www.coremedia.com/en/customers/iconic-brands
But beware: not all the main websites, that are linked here, also work with the CoreMedia CMS. Sometimes there are subsideries or special sites that work with it, or it provides the content to an ecommerce front-end.
The CoreMedia CMS is hard to detect because it is framework agnostic and hides itself very well. Then there are at least versions 7-10 in production with different characteristics.
I did the best I can to detect it anyway.


greenoid:
Ok. I see. I have to put < and > in the patterns. This is hard and will reduce the number of detected systems but I will try with a better version.

PR #2910, opened 2020-01-08
Added X-Powered-By: WP Engine header

adnanyousef:
In December 2019, WP Engine added the `X-Powered-By: WP Engine` response header and removed the `wpe-backend: *` header.

I left the old headers in but can remove them if preferred.


AliasIO:
Thanks!

PR #2916, opened 2020-01-17
Add autodeploy to DockerHub

abitrolly:


abitrolly:
@AliasIO if you add properly encrypt `DOCKER_PASSWORD`, the automatic Docker deployment should work.

AliasIO:
Good idea, thank you!

PR #2919, opened 2020-01-22
Removing old / unused applications

kyletaylored:
- Mypage (clone of Botble, causing issues with detection)
- MediaTomb - no longer in service
- CoinLab - no longer in service
- CloudCoins - no longer in service
- HP ProCurve - irrelevant
- Homeland - no longer in service
- HBase - not really detectable
- GoSquared - vague detection, at zero

Refers to #1432


AliasIO:
Thanks Kyle!

PR #2921, opened 2020-01-25
added IONOS MyWebsite CMS

hunsalz:
added IONOS MyWebsite CMS detection


hunsalz:
Closing PR - PR doesn't regard new and old MyWesite CMS version. I will provide an better solution soon.

PR #2928, opened 2020-02-05
October CMS detection fix

swiper:
The `October CMS` detection was not working due to a `=` in the cookie name. Now fixed.


AliasIO:
Thanks!

PR #2937, opened 2020-02-10
added Riskified

Johannes-Andersen:
Added riskified beacon to apps list. Example websites:

1. http://canadagoose.com/
2. https://www.debeers.co.uk/
3. https://www.danielwellington.com/uk/

https://www.riskified.com/ is a anti-fraud and anti revenue loss service for e-commerce platforms.


AliasIO:
Thanks!

PR #2938, opened 2020-02-11
Add Varbase CMS detection (a Drupal distribution)

moerazem:
Varbase is Drupal 8 distribution currently being used by over 1,100 sites. See https://www.drupal.org/project/varbase

Some example of sites using Varbase:
- https://detroitmi.gov/
- https://www.albawaba.com/
- http://balkans.aljazeera.net/
- https://www.unrwa.org/
- https://www.icarda.org/

The CMS uses Drupal essentially and always has "varbase_" in its markup or `<meta name="generator" content="Varbase" />`


AliasIO:
Thanks!

PR #2950, opened 2020-02-22
feat: add jsdom driver

Kikobeats:
Hello,

This PR introduces jsdom driver. I decided to create this driver because I want to wappalyzer passing all the necessary things as parameters

```js
const { url, html, headers } = await getHTML('https://kikobeats.com')
const wappalyzer = new Wappalyzer(url, { html, headers })
```

### DIFF
--- a/src/drivers/npm/browsers/jsdom.js
+++ b/src/drivers/npm/browsers/jsdom.js
@@ -1,0 +1,58 @@
+'use strict'
+
+const { Cookie } = require('tough-cookie')
+const jsdom = require('jsdom')
+
+const { JSDOM, VirtualConsole } = jsdom
+
+const uniq = arr => [...new Set(arr)]
+
+const parseCookie = str => Cookie.parse(str).toJSON()
+
+const getCookies = str =>
+  []
+    .concat(str)
+    .map(parseCookie)
+    .map(({ key: name, ...props }) => ({ name, ...props }))
+
+function Browser (opts) {
+  if (!(this instanceof Browser)) return new Browser(opts)
+
+  const { url, html, statusCode = 200, headers } = opts
+
+  const userAgent = headers['user-agent'] || null
+  const contentType = headers['content-type'] || null
+  const dom = new JSDOM(html, {
+    url,
+    runScripts: 'dangerously',
+    virtualConsole: new VirtualConsole()
+  })
+
+  const cookies = getCookies(headers['set-cookie'])
+
+  return {
+    visit: () => {},
+    userAgent,
+    cookies,
+    html,
+    statusCode,
+    contentType,
+    document: dom.window.document,
+    window: dom.window,
+    headers: {},
+    js: dom.window,
+    scripts: uniq(
+      Array.prototype.slice
+        .apply(dom.window.document.scripts)
+        .filter(({ src }) => src)
+        .map(({ src }) => src)
+    ),
+    links: uniq(
+      Array.from(dom.window.document.getElementsByTagName('a')).map(
+        ({ href }) => href
+      )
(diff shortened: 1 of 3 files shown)


src/wappalyzer.js
```diff
@@ -607,7 +607,7 @@ class Wappalyzer {
           headerName = headerName.toLowerCase();
 
           if (headerName in headers) {
-            headers[headerName].forEach((headerValue) => {
+            [].concat(headers[headerName]).compact(Boolean).forEach((headerValue) => {
```
Kikobeats: @AliasIO not sure why the header value needs to be an array.

Normally, headers is an object like this:

```js
const headers = {
  date: 'Sat, 22 Feb 2020 16:50:20 GMT',
  server: 'cloudflare'
}
```

src/drivers/npm/package.json
```diff
@@ -24,6 +24,8 @@
     "wappalyzer": "./cli.js"
   },
   "dependencies": {
+    "jsdom": "~16.2.0",
+    "tough-cookie": "~3.0.1",
     "zombie": "^6.1.2"
```
Kikobeats: I recommend you ship every driver in a different wappalyzer, something like:

-  @wappalyzer/jsdom
-  @wappalyzer/puppeteer
-  @wappalyzer/zombie

Using this approach then you only install what you need

src/drivers/npm/browsers/jsdom.js
```diff
@@ -0,0 +1,58 @@
...
+
+const uniq = arr => [...new Set(arr)]
+
+const parseCookie = str => Cookie.parse(str).toJSON()
+
+const getCookies = str =>
+  []
+    .concat(str)
+    .map(parseCookie)
+    .map(({ key: name, ...props }) => ({ name, ...props }))
```
Kikobeats: @AliasIO not sure why your implementation is using `name` instead of `key`

AliasIO:
The Zombie.js driver is built on jsdom, is there a benefit to using this?

src/wappalyzer.js
AliasIO: It's to handle duplicate headers.

src/drivers/npm/package.json
AliasIO: Yes, I might do this at some point.

Kikobeats:
I never used before, but looks like zombie performs a network call for getting the html content:
https://github.com/assaf/zombie#browservisiturl-options-callback

while in this implementation, all the things are passed via parameters, no internal network call will be performed.

In fact, if zombie is using jsdom as a dependency that's a good thing since the dependency will be deduped.

Kikobeats:
@AliasIO 

The point with this driver is to not do an internal fetch of the data.

I feel this PR goal could be achieved adding support to pass your own data:

- at puppeter: if `html` is passed, set it using [page.setContent](https://github.com/puppeteer/puppeteer/blob/v2.1.1/docs/api.md#pagesetcontenthtml-options) and don't performing the goto call.

- at zombie: if `html` is passed, set is using [browser.load](https://github.com/assaf/zombie#browserloadhtml-callback), skipping `browser.visit` action.

That could be a more reasonable approach since I really don't think more driver can be added.

Thoughts?

AliasIO:
I prefer the idea of separate packages, rather than officially supporting multiple browsers. You should already be able to create you own browser package and pull in Wappalyzer as a dependency. When I find the time I'll split the Zombie.js and Puppeteer browsers into their own package.

Kikobeats:
Closed, I created https://github.com/Kikobeats/simple-wappalyzer instead 🙂

PR #2951, opened 2020-02-23
Remove Implies Angular for Ionic

dmartinjs:
Ionic moved from Angular component to [Web Component](https://ionicframework.com/#agnostic),
so Ionic app can be build with Angular, React, [Stencil](https://stenciljs.com/) (Web Component compiler) and soon Vuejs.


AliasIO:
Thanks!

PR #2985, opened 2020-03-21
Added Miva detection

pharika31:
Added as Category 6 as its an ECP. In some cases, the version cannot be detected until specific pages like cart page or product page.

**Key logic:**

1. Lot of these sites have inline JS with variables MivaVM_API, MivaVM_Version.
2. Some have external JS in their HTML "mvga.js"
3. We can also see response header "content-disposition":filename=mvsg.js or MivaEvents.js

Here are some example websites that the solution was tested on:
- https://www.directhomemedical.com/
- https://www.lpgear.com/
- https://www.graingermckoy.com/
- https://www.christmastraditions.com/
- https://www.littmanbros.com/
- https://www.dieselpowerproducts.com/
- https://www.thepaperstore.com/
- https://www.thethreadexchange.com/
- https://www.bedbathhome.com/
- https://www.kitchencabinetdepot.com/

There are multiple other examples listed here:
https://pro.builtwith.com/report/list/7e94b3ae-7eda-4fd5-bc25-eaa6bda0b224


AliasIO:
Thanks!

PR #2988, opened 2020-03-22
Update the regex of the Airform.io

cjpatoilo:
Huge thanks to @castrolol
You're amazing! 🚀


src/apps.json
```diff
@@ -554,7 +554,7 @@
         61
       ],
       "html": [
-        "<form[^>]+?action=\"[^\"]+airform.io"
+        "\<form(.*?)action=['"](([a-z0-9]*?:?(\/\/)?airform\.com)[^ '"]*?)['"]"
```
AliasIO, line 557: This doesn't look right, what's wrong with the existing pattern?

cjpatoilo: Please, consider this test:
https://regex101.com/r/R7vWj9/5

@AliasIO Let me know what went wrong for you.

kingthorin: The groups should be non-capturing.

You didn't explain what's wrong with the existing pattern.

Making two small modifications to the existing pattern makes all the matches in your regex101 test. `<form[^>]+?action=\"[^\"]*airform\.(?:io|com)` in 313 steps, vs 453 for your regex.

cjpatoilo: Sorry guys. I will fix the domain extension now.

cjpatoilo:
@AliasIO @kingthorin 

Please, consider seeing my test `<form[^>]+?action=\"[^\"]*airform\.io[^>]+?>`: https://regex101.com/r/R7vWj9/7

Let me know if have there is something wrong.

kingthorin:
Seems fine. Is there an announcement somewhere about the `.com` to `.io` switch?

It could also be simplified a bit to match less and be more forgiving about spaces in HTML syntax: https://regex101.com/r/E1dTJJ/3

cjpatoilo:
@kingthorin 

> Seems fine. Is there an announcement somewhere about the `.com` to `.io` switch?

Always `.io`. 

> It could also be simplified a bit to match less and be more forgiving about spaces in HTML syntax: https://regex101.com/r/E1dTJJ/3

Your example enables a possible false positive. For example: `<a action="airform.io/">Somthing</a>`

kingthorin:
Thanks, it wasn't clear that the original was wrong, and fair enough. LGTM, for whatever that's worth 🤷‍♂️ 😉

cjpatoilo:
@AliasIO @kingthorin

Please, consider seeing my test:

before file changes https://regex101.com/r/R7vWj9/9
after file changes https://regex101.com/r/R7vWj9/10

Let me know if have there is something wrong.

src/apps.json
```diff
@@ -562,7 +562,7 @@
         61
       ],
       "html": [
-        "<form[^>]+?action=\"[^\"]+airform\\.io"
+        "<form[^>]+?action=\"[^\"]*airform\.io[^>]+?>"
```
AliasIO, line 565: The period needs to be escaped twice. I'll fix this up.

AliasIO:
Thanks!

src/apps.json
cjpatoilo: @AliasIO huge thanks!

cjpatoilo:
@AliasIO 

See, It doesn't show Airform when Wappalyzer is running https://airform.io

Maybe it needs to check the JS files as well. Is it possible that I am right? 🤔

AliasIO:
That page doesn't seem to have the HTML that's being matched, you may need to test it on a website that's actually using Airform.

cjpatoilo:
@AliasIO 

Exactly! This is my question. For example, I use Airform on https://sponsordeck.co
But I use React too. So this form (with Airform) is rendered by React (JavaScript).

Look:

image

Of course, that code source does not show the form with Airform:

image

So I think I will need to extend the Wappalyzer support for JavaScript files, right? 🤔

AliasIO:
I see what you mean. Wappalyzer doesn't look for dynamic DOM changes although this may change in the future. Bundled modules can't really be identified at the moment.

PR #2989, opened 2020-03-22
Add VHosting Solution as PaaS

samsteve169:
Add VHosting Application
Add VHosting Logo


AliasIO:
This is a web host, not PaaS.

PR #3008, opened 2020-04-03
NivCMS remove empty elements

kingthorin:
Having empty elements just results in unnecessary parsing or incorrect matching.

NivCMS was added in:
https://github.com/AliasIO/wappalyzer/pull/2995/files

Perhaps this is something that could be considered further to https://github.com/AliasIO/wappalyzer/issues/2896 (which mentions linting or unittesting version syntax, however, linting for json format either consistently including all elements or not including empty elements might be a good move too.)

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -11786,16 +11786,12 @@
         "cookie_name": "NivCMSPro"
       },
       "icon": "nivsoft.png",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #3011, opened 2020-04-05
Google Shield Project

### LINKED ISSUE #3007: Google Shield - DDoS protection, like Cloudflare for journalist
Link:
https://projectshield.withgoogle.com/landing

Test by checking header and this if exist:
x-shield-request-id

Examples:
https://krebsonsecurity.com/
https://www.thelocal.com/
https://eclecticrhetoric.com/about/

Triloworld:
Resolve #3007

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -4420,6 +4420,17 @@
       "script": "apis\\.google\\.com/js/[a-z]*\\.js",
       "website": "http://plus.google.com"
     },
+    "Google Shield Project": {
+      "cats": [
+        63
+      ],
+      "headers": {
+        "via": "^([0-9]\\.[0-9])\\;version:\\1 google",
+        "x-shield-request-id": ""
+      },
+      "icon": "Google Shield Project.svg",
+      "website": "https://projectshield.withgoogle.com"
+    },
     "Google Sites": {
       "cats": [
         1
--- a/src/apps.json
+++ b/src/apps.json
@@ -4425,7 +4425,7 @@
         63
       ],
       "headers": {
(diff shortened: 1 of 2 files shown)


src/apps.json
```diff
@@ -4420,6 +4420,17 @@
       "script": "apis\\.google\\.com/js/[a-z]*\\.js",
       "website": "http://plus.google.com"
     },
+    "Google Shield Project": {
+      "cats": [
+        63
+      ],
+      "headers": {
+        "via": "^([0-9]\\.[0-9])\\;version:\\1",
```
AliasIO, line 4428: This way GSP will be identified on any request with a `via` header. You could change it to `^([0-9]\\.[0-9])\\;version:\\1;confidence:0` to only use it for version detection. Is there nothing else in the `via` header?

Triloworld: If I add another set for header, it is used in "AND" context of two of them? or "OR"?
It need to be send in header "x-shield-request-id". Don't find example with 2 header to be sure how it work in real life scenario

src/apps.json
```diff
@@ -4420,6 +4420,17 @@
...
+        "x-shield-request-id": ""
```
Triloworld, line 4429: @AliasIO This is required to show that Google Shield Project is in use.

src/apps.json
AliasIO: It's OR. If you add `\\;confidence:0` it can still be used for version detection but the `via` header alone won't trigger a detection. A lot of sites that don't use GSP use that header, including [wappalyzer.com](https://www.wappalyzer.com).

PR #3019, opened 2020-04-07
fixed IONOS icon

hunsalz:
added an IONOS icon with a better aspect ratio to fit layout


AliasIO:
Why are the two versions of MyWebsite now separate technologies? They should be combined with version detection.

hunsalz:
Both versions have nothing in common than the marketing name. Version 8 is self developed by IONOS and available for existing customers only. Version 9 is based on Duda and actively sold. That's the reason for the differentiation.

If I get you right you suggest to combine them in one ruleset? - I will try, but the ruleset is probably an ugly regex mess.

AliasIO:
If the website and icon can be the same you can just add the new JS pattern the existing section. You don't need to combine the regular expressions into one.

PR #3023, opened 2020-04-07
improved WHMCS detection

hunsalz:
fix for https://github.com/AliasIO/wappalyzer/issues/2940#issuecomment-589302219

improved WHMCS detection based on js detection instead of cookies


AliasIO:
Thanks!

PR #3033, opened 2020-04-18
Change Typecho to small icon

scvoet:


AliasIO:
The icon should be square and preferably SVG.

PR #3043, opened 2020-04-21
ZEIT changed its name to Vercel

spVierSechs:
(The company name and the product name are now unified, which makes the "Now" naming obsolete too.)  
  
`x-now-trace` seems to be unchanged.  
  
The SVG is taken from [the official page](https://vercel.com/design/vercel) and the only difference seems to be the fact that the very subtle gradient was removed.


AliasIO:
Thanks!

PR #3049, opened 2020-04-26
Add icon & detection for MyCashflow

villeristi:
Adds detection for a finnish webstore-provider [MyCashflow](https://www.mycashflow.fi/).


AliasIO:
Thanks!

PR #3052, opened 2020-04-28
Added J2Store as an Ecommerce Platform

MarcDillar:
J2Store is a Joomla extension.

Website: https://www.j2store.org/

Examples:
http://www.guerrilladrone.com/
https://knotpredictable.com/
https://www.lavalcosmetics.co.uk/
https://www.mipotty.com/


AliasIO:
Thanks!

src/apps.json
```diff
@@ -5257,6 +5257,15 @@
       "icon": "Ionicons.png",
       "website": "http://ionicons.com"
     },
+    "J2Store": {
+      "cats": [
+        6
+      ],
+      "html": "var j2storeURL",
```
AliasIO, line 5264: Sorry, I had to undo the merge because this pattern is invalid. It should contain at least HTML opening tag or it will match any text.

PR #3057, opened 2020-05-04
Add AppDynamics detection

nirlya:
**AppDynamics**

AppDynamics is an application performance management (APM) and IT operations analytics (ITOA) company based in San Francisco. The company focuses on managing the performance and availability of applications across cloud computing environments as well as inside the data center. In March 2017, AppDynamics was acquired by Cisco for $3.7 billion.

It is a Gartner Leader of APM like Dynatrace & New Relic (support already existing) :

**Regex :**
image

Website : https://www.appdynamics.com


AliasIO:
Thanks!

PR #3086, opened 2020-05-17
Add Tailwind CSS identifier

imtrent:
**Regex Expression**
`[^>]*class=\"[^\"]*(?:sm:|md:|lg:|xl:)`

The regex expression detects if there are any class names starting with either "sm:", "md:", "lg:", or "xl:" within the HTML. 

These class name prefixes are heavily used in websites using [Tailwind CSS](https://tailwindcss.com/), as they denote breakpoint styles.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -10747,7 +10747,10 @@
       "cats": [
         66
       ],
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

bharadr91:
Looks like this feature is not implemented yet.

PR #3092, opened 2020-05-20
(Huge PR) Make the logo collection storage size smaller

Stryxus:
This all begun with me wanting to make an app which can convert and resize a mass of images in one go, this idea turned into TooManyPixels, my little and simple app (Which is open source and downloadable on the repo). After its planning, i decided Wappalyzer would be a good testing bed because of its huge collection of images and if i accidentally delete the source folder, i could easily re-clone.

When i was half way through making TooManyPixels, i realised, i could actually fix the mess of different sizes which Wappalyzer is burdened with because it would take forever to sift through every image and correct it, this also led to the discovery that Wappalyzers logo collection storage size could be lowered by 30.14%!

I have left the old /icons/ folder just in case anyone wants to compare. The replacement is /icon/

- All images in this PR are 16x16 pixels (I dont see any use case of this being any higher although i can, at any time, convert all files to 32x32 or higher or lower)
- The size reduction is 30.14% at a compression rate of 50%
- All images which were not of equal size ratio are now equal

Since i forgot committing changes adds onto existing PR's and you cant make a new PR, iv also removed ASP.NET implying IIS server. ASP.NET projects no longer depend on IIS and can be run standalone, on NGINX, Apache or others.


AliasIO:
The raster images should be at least 32x32 for retina displays but ideally we have vector images for everything. I'm seeing GIF and BMP files in your pull request, are you sure those are smaller and the compression is lossless?

Stryxus:
I can change it to 32x32 and remove the BMP and GIF files right now. Quite interesting that these showed up somehow. Oh well, at least im aware that it has to be fixed.

Stryxus:
Alright, all files are 32x32 and all BMP's and GIF's are gone.

Before: 2.37MB
After: 1.75MB at a 26.27% reduction.

AliasIO:
It's hard to compare what's actually changed because you copied the folder. The images should be either SVG or PNG. PNGs are already being compressed. I know there are some PNG files that are too large, resizing those to 32x32 is useful. Think you can limit the PR to that? Let me know if I'm missing something.

Stryxus:
Ye i can. Ill do it today.

Stryxus:
Latest results:

- Updated fork with latest project structure.
- 136 oversized images have been resized and recompressed accordingly.
- 2.41MB to 1.76MB. A 26.86% reduction.

Stryxus:
Im going to create another pull request with the logos. There are probably quite a few conflicts by now and this pull request contains changes which are no longer compatible.

PR #3098, opened 2020-05-21
Confidence level is not calculated for implied apps

andywaite:
Issue: Implied apps always have confidence level 0. This does not seem like desired behaviour. If e.g. WordPress is detected with 100 confidence then PHP should also surely be detected with 100 confidence.

Looking into it, the function `app.getConfidence()` was being called BEFORE resolving implied apps. It never got called on the detected implied apps. I figured this must be an error, and it needs to be called again once the implied apps have been detected.

It's kinda messy running this twice on most apps and there may be a more elegant solution.

### DIFF
--- a/src/wappalyzer.js
+++ b/src/wappalyzer.js
@@ -237,6 +237,15 @@ class Wappalyzer {
       resolveExcludes(apps, this.detected[url]);
       this.resolveImplies(apps, url.canonical);
 
+      // Resolve confidence again, required for implied apps
+      Object.keys(apps).forEach((appName) => {
+        const app = apps[appName];
+
+        if (!app.getConfidence()) {
+          delete apps[app.name];
+        }
+      });
+
       this.cacheDetectedApps(apps, url.canonical);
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks, I'll have a look into this.

AliasIO:
I just tested it by setting the confidence for Nuxt.js to 10%, and the implied technologies inherit this correctly. There was an issue with the confidence not being displayed at all, which I fixed.

Did you notice this in the browser extension or NPM module?

Screen Shot 2020-05-28 at 10 58 38 am

andywaite:
> I just tested it by setting the confidence for Nuxt.js to 30%, and the implied technologies inherit this correctly. There was an issue with the confidence not being displayed at all, which I fixed.
> 
> Did you notice this in the browser extension or NPM module?
> 
> Screen Shot 2020-05-28 at 10 58 38 am

This was using the NPM module

AliasIO:
This has been fixed.

PR #3056, opened 2020-05-03
Alpine js

repat:
-- https://github.com/alpinejs/alpine

_Alpine.js_ is the foundation of the popular [Laravel Livewire](https://laravel-livewire.com) package but can also be used as a standalone library for simple things like modals or dropdowns where you don't necessarily have to pull in Vue or React. The syntax is very similar to Vue but it works without a virtual DOM. It's relatively new but is getting popular as part of the TALL/TTALL Stack: Tailwind, (Turbolinks,) Alpine, Laravel, Livewire

⭐ 7.1k stars on GitHub

#### Examples
* https://novapackages.com
* https://www.alpinetoolbox.com
* https://tallstack.dev
* https://laravel-livewire.com
* Curated List of resources: https://github.com/alpinejs/awesome-alpine
* Used-By GitHub Feature: https://github.com/alpinejs/alpine/network/dependents?package_id=UGFja2FnZS03NjA0NzgxMzA%3D

#### Regex
image

#### Icon
64x64 PNG

image

PR #3104, opened 2020-05-23
Improve Alpine.js detection

adevade:
The regex pattern from PR #3056 would match `<div x-data>`, but it would also match `<div lightbox-data>` and gave a lot of false positives. 

This PR fixes that problem, lowers the confidence slightly for html match and adds additional methods of identification.

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -83,8 +83,14 @@
       "cats": [
         12
       ],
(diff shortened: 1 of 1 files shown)


adevade:
**HTML REGEX BEFORE**


**HTML REGEX AFTER**

AliasIO:
Thanks!

PR #3105, opened 2020-05-23
Update Vue.js metadata

adevade:
- **ElementUI** was implying `Vue` instead of `Vue.js`.
- Vue's website link was not using `https`.


AliasIO:
Thanks!

PR #3106, opened 2020-05-23
Always use string for implies with only one item

adevade:
Changes all `implies` arrays that only had one item to use string syntax instead. 

If you'd prefer to always use array, just let me know and I'll update the PR.


AliasIO:
Thanks!

PR #3135, opened 2020-05-30
Add Redis (database) and Redis Object Cache (WP plugin)

tillkruss:
This PR adds [Redis](https://redis.io) as database as well as detection for the [Redis Object Cache](https://wordpress.org/plugins/redis-cache/) plugin for WordPress (used by 60k+ sites).

Redis itself has 43k stars.
- https://wordpress.org/plugins/redis-cache/

Redis Object Cache only has ~160 stars, but it's _active installations base_ is 60,000+ according to wordpress.org.

- https://github.com/tillkruss/redis-cache
- https://wordpress.org/plugins/redis-cache/


AliasIO:
Thanks!

PR #3139, opened 2020-06-01
Pattern tweaks

kingthorin:
Per recent CI runs, ex: https://travis-ci.org/github/AliasIO/wappalyzer/builds/692417278#L264-L320

#### One

There were two in this group that I didn't tackle:
```text
Suspicious period (".") in pattern. Should this have been escaped?
	App: GeneXus
	Pattern: script: .*/static/gxgral.js
Suspicious period (".") in pattern. Should this have been escaped?
	App: GeneXus
	Pattern: script: .*/static/gxtimezone.js
```

They appear to be valid periods, however is it really necessary to start a pattern with wildcard (`.*`)? If it isn't necessary I'd be glad to adjust them as well.

Edit: The GeneXus patterns were also fixed in the end, per additional comments.

#### Two

When using the patterns in Java the following "illegal repetition" issues occur, because of unmatched curly braces. According to regex101.com escaping these doesn't impact the JavaScript behavior:

```text
java.util.regex.PatternSyntaxException: Illegal repetition near index 39
<script[^>]*>\s*Raven\.config\('[^']*', {\s+release: '([0-9\.]+)'

java.util.regex.PatternSyntaxException: Illegal repetition near index 26
<script>(?:.|\n)*"refresh":{"version":"([\w\.]+)\s
```

Signed-off-by: Kingthorin <kingthorin@users.noreply.github.com>

### DIFF
--- a/src/apps.json
+++ b/src/apps.json
@@ -991,7 +991,7 @@
         10
       ],
       "icon": "AppDynamics.png",
-      "script": "adrum\\.js|adrum.([0-9].*)\\.js\\;version:\\1",
+      "script": "adrum\\.js|adrum\\.([0-9].*)\\.js\\;version:\\1",
       "website": "https://appdynamics.com"
     },
     "AppNexus": {
@@ -3922,7 +3922,7 @@
       "cats": [
         24
       ],
-      "html": "/<!--\\s*\\*\\s*FreeTextBox v\\d+ \\(([.\\d]+)(?:(?:.|\\n)+?<!--\\s*\\*\\s*License Type: (Distribution|Professional)License)?/i\\;version:\\1 \\2",
+      "html": "/<!--\\s*\\*\\s*FreeTextBox v\\d+ \\(([.\\d]+)(?:(?:\\.|\\n)+?<!--\\s*\\*\\s*License Type: (Distribution|Professional)License)?/i\\;version:\\1 \\2",
       "icon": "FreeTextBox.png",
       "implies": "Microsoft ASP.NET",
       "js": {
@@ -4191,7 +4191,7 @@
       ],
       "icon": "GitBook.png",
       "meta": {
-        "generator": "GitBook(?:.([\\d.]+))?\\;version:\\1"
+        "generator": "GitBook(?:\\.([\\d.]+))?\\;version:\\1"
       },
       "url": "^https?://[^/]+\\.gitbook\\.com/",
       "website": "https://www.gitbook.com"
@@ -4972,7 +4972,7 @@
       "cats": [
         57
       ],
-      "html": "powered by <a [^>]*href=\"http://hugo.spf13.com",
+      "html": "powered by <a [^>]*href=\"http://hugo\\.spf13\\.com",
       "icon": "Hugo.png",
       "meta": {
         "generator": "Hugo ([\\d.]+)?\\;version:\\1"
@@ -5489,7 +5489,7 @@
         18
       ],
       "headers": {
-        "X-Powered-By": "Servlet(?:.([\\d.]+))?\\;version:\\1"
+        "X-Powered-By": "Servlet(?:\\.([\\d.]+))?\\;version:\\1"
       },
       "icon": "Java.png",
       "implies": "Java",
@@ -6427,7 +6427,7 @@
         "mivaJS.Screen": ""
       },
       "headers": {
-        "content-disposition": "filename=(?:mvga.js|MivaEvents.js)"
+        "content-disposition": "filename=(?:mvga\\.js|MivaEvents\\.js)"
       },
       "website": "http://www.miva.com"
     },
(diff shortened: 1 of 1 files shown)


AliasIO:
> however is it really necessary to start a pattern with wildcard

No, it's not necessary.

kingthorin:
Thanks @AliasIO all fixed up.

src/apps.json
```diff
@@ -3922,7 +3922,7 @@
       "cats": [
         24
       ],
-      "html": "/<!--\\s*\\*\\s*FreeTextBox v\\d+ \\(([.\\d]+)(?:(?:.|\\n)+?<!--\\s*\\*\\s*License Type: (Distribution|Professional)License)?/i\\;version:\\1 \\2",
+      "html": "/<!--\\s*\\*\\s*FreeTextBox v\\d+ \\(([.\\d]+)(?:(?:\\.|\\n)+?<!--\\s*\\*\\s*License Type: (Distribution|Professional)License)?/i\\;version:\\1 \\2",
```
kingthorin, line 3925: Looking at view-source:http://www.freetextbox.com/demos/ I think this one was actually meant to be wildcard (`.+` or newline+).

src/apps.json
```diff
@@ -4191,7 +4191,7 @@
       ],
       "icon": "GitBook.png",
       "meta": {
-        "generator": "GitBook(?:.([\\d.]+))?\\;version:\\1"
+        "generator": "GitBook(?:\\.([\\d.]+))?\\;version:\\1"
```
kingthorin, line 4194: I found a few samples like `meta name="generator" content="GitBook 3.2.2"` so this isn't literal period, however, it probably should be any single char either, it should likely be `\s`?

Also this is a pre-existing shortcoming but this would miss things like `<meta name="generator" content="bookdown 0.17 and GitBook 2.6.7" />` which I also found (but the original would have as well).

src/apps.json
```diff
@@ -12180,7 +12180,7 @@
       ],
       "headers": {
         "Server": "Winstone Servlet (?:Container|Engine) v?([\\d.]+)?\\;version:\\1",
-        "X-Powered-By": "Winstone(?:.([\\d.]+))?\\;version:\\1"
+        "X-Powered-By": "Winstone(?:\\.([\\d.]+))?\\;version:\\1"
```
kingthorin, line 12183: This one also probably isn't literal period. The few examples I was able to find via shodan were like `X-Powered-By: Servlet/2.5 (Winstone/1.0.5)`

While the original pattern does work it should probably be an escaped slash instead of period. ex: `Winstone(?:\/([\d.]+))?`

src/apps.json
```diff
@@ -12857,7 +12857,7 @@
       "cookies": {
         "ZM_TEST": "true"
       },
-      "html": "<script>(?:.|\\n)*\"refresh\":{\"version\":\"([\\w\\.]+)\\s\\;version:\\1",
+      "html": "<script>(?:\\.|\\n)*\"refresh\":\\{\"version\":\"([\\w\\.]+)\\s\\;version:\\1",
```
kingthorin, line 12860: The first one change here might be wrong, I could find any examples to check against. I guess it's likely mean to be script tag followed by any number of characters or newlines then refresh in quotes.

src/apps.json
```diff
@@ -5489,7 +5489,7 @@
         18
       ],
       "headers": {
-        "X-Powered-By": "Servlet(?:.([\\d.]+))?\\;version:\\1"
+        "X-Powered-By": "Servlet(?:\\.([\\d.]+))?\\;version:\\1"
```
kingthorin, line 5492: This one also probably isn't literal period. The few examples I was able to find via Shodan were like `X-Powered-By: Servlet/2.4` (or the winstone one)

While the original pattern does work it should probably be an escaped slash instead of period. ex: `Servlet(?:\/([\d.]+))?`

kingthorin:
@AliasIO I've gone through and comment on a number of these where I may have made mistakes. In the majority of cases I've suggested alternatives. Please let me know what you think.

AliasIO:
Thank you!

kingthorin:
Uh, @AliasIO did you read through the comments/questions? Are you sure I didn't break something?

I'm glad to open a followup PR if something needs further tweaking.

AliasIO:
Ah sorry, I did not. I'll have a look at it when I get a chance.

kingthorin:
No problem, just let me know.

kingthorin:
@AliasIO I'd really like to fix these if I did indeed not make the correct choices on them. Please have a look through the comments/questions above and let me know. I'll happily submit a fix PR.

AliasIO:
I fixed these up. The FreeTextBox pattern was pretty bad, HTML patterns probably shouldn't have wildcards that match text of arbitrary length. I'll have a think how we can catch these with tests.

kingthorin:
Related fixes seem to be in: https://github.com/AliasIO/wappalyzer/commit/fd58b18a9551882cd057ab3fe789d69973d5a108#diff-557d7d60c04092a680ce152f367618c0 :+1:

PR #3167, opened 2020-06-11
Add RebelMouse specific header detection

kyletaylored:
RebelMouse, a WordPress-like CMS, uses specific caching headers for their service which can be detected.


AliasIO:
Thanks!

PR #3171, opened 2020-06-11
Add JSDoc blocks to all functions

kyletaylored:
While looking into another issue, I found it difficult to parse functions and understand their purpose. One step in the right direction would be add JSDoc blocks to all functions and Object methods to help onboard new contributing developers improve overall readability.

- Focused on Wappalyzer and web extension
- iframe.js hasn’t been cleaned / formatted in awhile, so it looks like a lot of changes, but is just JSDoc blocks and formatting.

### DIFF
--- a/src/drivers/webextension/js/content.js
+++ b/src/drivers/webextension/js/content.js
@@ -3,6 +3,9 @@
 /* globals chrome */
 
 const Content = {
+  /**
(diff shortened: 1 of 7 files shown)


AliasIO:
Thank you, I'll go over it now.

PR #3176, opened 2020-06-12
Fix overly broad Angular detection

lillesvin:
Was triggered on "rectangular[several KiB of minified script].js"
Now requires a word boundary before "angular" and allows max. 32
characters between that and ".js"


lillesvin:
Ok, so Travis fails but it doesn't seem related to my change at all..?

By the way, the warning about the unescaped "." in my regex can be ignored. It is indeed on purpose.

AliasIO:
Thanks!

Issue #3181, opened 2020-06-14
npm driver - infinite recursion causes memory leak / crash


ThomasOrlita:
I was analyzing some sites and found [this one](https://www.hespress.com/), where wappalyzer would get stuck – the chromium process would be taking all the CPU and memory.
After some debugging, I found the root cause to be in the [`getJs`](https://github.com/AliasIO/wappalyzer/blob/master/src/drivers/npm/driver.js#L60) function, where it is dereferencing the `window` object, and because the site has added more custom functions/objects to `Object`'s prototype, the function got stuck, recursively adding a huge number of properties on all objects...

An easy fix would be to add `hasOwnProperty` check to line [79](https://github.com/AliasIO/wappalyzer/blob/master/src/drivers/npm/driver.js#L79):

    if(!obj.hasOwnProperty(key)) return;

I checked a few other sites with this change and it worked well, but it's possible it might break detection for some other technologies.

Another option might be using for example [JSON.prune](https://github.com/Canop/JSON.prune/) package, or adding a timeout for evaluating the script.

Or, if possible, evaluate `processJs` (or the part of the function that accesses `window`) directly on the page where it has access to `window`.

**Platform**
* NPM

AliasIO:
Fixed in 088590b0661029138ac90950df988001d2f53e81 and v6.0.10, thank you!

AliasIO:
You're right about moving the evaluation onto the page, not sure why I didn't do it like that in the first place. Fixed again in v6.0.12.

PR #3184, opened 2020-06-15
npm - add timeout to javascript evaluation

ThomasOrlita:
Just stumbled upon a similar issue to #3181: [this site](http://www.caixa.gov.br/) exposes window.Module, which for some reason contains `HEAP` properties with millions of items – again causing the `dereference` function to execute until it runs out of memory...

I added a timeout for the javascript execution to stop if the browser doesn't return within `maxWait` seconds.

### DIFF
--- a/src/drivers/npm/driver.js
+++ b/src/drivers/npm/driver.js
@@ -375,7 +375,12 @@ class Site {
     ).jsonValue()
 
     // JavaScript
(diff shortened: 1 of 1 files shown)


AliasIO:
This should no longer be an issue now that the evaluation has been moved onto the page.

PR #3227, opened 2020-07-18
New category - Cookie compliance

rockeynebhwani:
@AliasIO  - Addded cookie compliance category and added initial list of vendors


AliasIO:
Thank you!

PR #3234, opened 2020-07-21
Added detection for WebAR / More 'Cookie Compliance Vendors' / Add few eCommerce platforms and a generic 'cart functionality' detection / Renamed category 16 to Security and updated vendors in this category

rockeynebhwani:
https://github.com/AliasIO/wappalyzer/issues/3216


src/apps.json
```diff
@@ -12213,6 +12213,19 @@
       "script": "web2py\\.js",
       "website": "http://web2py.com"
     },
+    "WebAR": {
+      "cats": [
+        19
+      ],
+      "html": [
+        "<model-viewer",
+        "\\.usdz",
```
AliasIO, line 12222: HTML patterns should always match at least one opening or closing tag to prevent it from matching plain text.

rockeynebhwani: @AliasIO - Intention was her to match <model-viewer tag OR presence of files with .usdz OR .glb extension.. Is there something else I need to use for matching file patterns like this?

src/apps.json
```diff
@@ -2479,6 +2491,34 @@
       "script": "consent\\.cookiebot\\.com",    
       "icon": "Cookiebot.svg",
       "website": "http://www.cookiebot.com"
+    },
+    "CookieConsent": {
+      "cats": [
+        67
+      ],
+      "script": [
+          "www\\.cookieconsent\\.com/releases/*/cookie-consent\\.js"
```
AliasIO, line 2500: The `*` makes the `/` in `releases/` optional, that's probably not what you intended?

src/apps.json
AliasIO: If the files are being included with a `link` or `script` tag you can use that but Wappalyzer doesn't check if a file exists on the server.

rockeynebhwani: Actually, I can remove check for .usdz and .glb as implementation wise, it looks like this on a webpage

`<model-viewer src="shared-assets/models/Astronaut.glb" alt="A 3D model of an astronaut" auto-rotate camera-controls></model-viewer>`

src/apps.json
rockeynebhwani: you are right... I don't want `/` in `releases/` to be optiona.
I tried
`www\\.cookieconsent\\.com/releases/([\\d.]+)/cookie-consent\\.js\\;version:\\1`
and
`www\\.cookieconsent\\.com/releases/.*/cookie-consent\\.js`

but none seems to do the trick... As you can see my regex skills are not great, can you tell me what am I dong wrong here?

src/apps.json
rockeynebhwani: @AliasIO - please see https://github.com/AliasIO/wappalyzer/pull/3234/commits/2a1985f01d007051cfba93936f03798c61f42a7f and let me know if it's OK now.

rockeynebhwani: sorry.. realized that I added a test file by mistake. removed that in https://github.com/AliasIO/wappalyzer/pull/3234/commits/c3ef881271570007a075a40a3149ca70693e1c1d

src/apps.json
AliasIO: Either look fine if you're just trying to match digits.

rockeynebhwani: ok.. not sure why it's not being picked during testing in that case. I will have to look at this further.

rockeynebhwani: @AliasIO - only example site, I have for this is https://www.cookieconsent.com/ and I see this JS request being made 
image

rockeynebhwani: @AliasIO - I have removed CookieConsenet from apps.json for now as this site is triggering .js file from a iframe (I am not sure if a JS file is triggered from iframe, we detect it). This may be the reason, my tests not working. I  have removed it for now as I  have no other example sites to test this. I also adjusted couple of other regexes based on your feedback. Please review - https://github.com/AliasIO/wappalyzer/pull/3234/commits/fe665d65b169d9b4f422b68d0c8efc4396784a0e

If you have time today to review and merge, I would like to get this PR closed and start a new one as this has grown big now

src/apps.json
```diff
@@ -1955,6 +1967,21 @@
       "script": "/cargo\\.",
       "website": "http://cargocollective.com"
     },
+    "Cart Functionality": {
+      "cats": [
+        6
+      ],
+      "html": [
+          "<a[^>]*href=.*/Cart",
```
AliasIO, line 1975: Please avoid `.*` in `html`. The can match very long strings that result in poor performance.

rockeynebhwani: @AliasIO - Please check if it's better now... changed it to` [^>]*`

If OK, please merge

src/apps.json
rockeynebhwani: @AliasIO - You happy to resolve this comment?

AliasIO:
Thanks!

PR #3239, opened 2020-07-22
Update WebSphere Commerce and SAP Commerce sites detection

jrharalson:
Slight shift in the regex pattern works much better. Pushing a PR with the update now. I'm using "a" "link" and "script" as they are references to the IBM url patterns that will drive higher detected then just the "a". Site will use "/wcsstore/" and/or "/webapp/wcs/" in depending on the Site Context.

"html": "(?:a|link|script)[^>](?:href|src)=".(/wcsstore/|webapp\/wcs)"


src/apps.json
```diff
@@ -5101,7 +5101,7 @@
         6
       ],
       "cpe": "cpe:/a:ibm:websphere_commerce_suite",
-      "html": "href=\"(?:\\/|[^>]+)webapp\\/wcs\\/",
+      "html": "(?:a|link|script)[^>]*(?:href|src)=\".*(\/wcsstore\/|webapp\\/wcs)",
```
AliasIO, line 5104: I just noticed that this pattern should have an opening tag so that it doesn't match plain text. This should work:

`<(?:a|link|script)[^>]*(?:href|src)=\".*(\/wcsstore\/|webapp\\/wcs)`

jrharalson: Good catch. Yup - I missed this in my check-in

AliasIO:
Thanks!

PR #3243, opened 2020-07-23
Added New Category - Accessibility

adityapandey1998:
Added a new Category - **Accessibility**
Some apps added in this category - 
- EqualWeb (https://www.equalweb.com/)
- UserWay (https://userway.org/)
- Texthelp (https://www.texthelp.com/en-gb/products/browsealoud/)
- Recite me (https://reciteme.com/)
- Essential Accessibility (https://www.essentialaccessibility.com/)
- uRemediate (https://www.user1st.com)
- Adally (https://adally.com/)
- FACILiti (https://ws.facil-iti.com/)
- accessiBe (https://accessibe.com/)
- audioeye (https://www.audioeye.com/)
- HikeOrders (https://hikeorders.com/)
- make sense (https://mk-sense.com/)

(Issue is given here - https://github.com/AliasIO/wappalyzer/issues/3228)


src/apps.json
```diff
@@ -1149,6 +1166,15 @@
       "icon": "Atlassian Statuspage.svg",
       "website": "https://www.statuspage.io/"
     },
+    "AudioEye": {
+      "cats": [
+        68
+      ],
+      "html": "<iframe.*audioeye\\.com/frame/cookieStorage",
```
AliasIO, line 1173: `.*` should be avoided in HTML patterns as matches can very long (bad for performance) and match plain text.

src/apps.json
```diff
@@ -3559,6 +3593,14 @@
       "icon": "syncfusion.svg",
       "website": "https://www.syncfusion.com/javascript-ui-controls"
     },
+    "EssentialAccessibility": {
+      "cats": [
+        68
+      ],
+      "html": "<a .*essentialaccessibility\\.com/.*\\?utm_source=.*utm_medium=.*",
```
AliasIO, line 3600: Avoid `.*`. The more specific the better but you can use `[^>]*`.

Avoid links as well. A link to a website is not a strong signal that a technology is in use.

src/apps.json
```diff
@@ -6654,6 +6712,15 @@
       ],
       "website": "http://mailchimp.com"
     },
+    "Make-Sense": {
+      "cats": [
+        68
+      ],
+      "html": "<script.*mk-sense\\.com/aweb\\?license",
```
AliasIO, line 6719: Avoid `.*`. Can we use `script` instead of `html`?

src/apps.json
```diff
@@ -9289,6 +9356,14 @@
       },
       "website": "http://www.rayo.ir"
     },
+    "Recite Me": {
+      "cats": [
+        68
+      ],
+      "html": ".*(api\\.reciteme\\.com/asset/js).*",
```
AliasIO, line 9363: Avoid `.*` and match at least one HTML tag.

rockeynebhwani: @AliasIO - For reciteme, code looks like this and I think that's the reason for generic match. Do you have ideas to deal with this?
image

src/apps.json
rockeynebhwani: @adityapandey1998 - you can try changing this to following `"html": "]+ alt=\"Essential Accessbility Logo",`

src/apps.json
AliasIO: In this case you could use `js` and check if `serviceUrl` has a value that matches `api.reciteme.com`.

src/apps.json
adityapandey1998: @rockeynebhwani The img tag isn't present in other sites. All sites using essential accessibility use this type of link - essentialaccessibility.com/<site_name>?utm_source=<>utm_medium=<>

rockeynebhwani: @adityapandey1998 - I understand.. I saw few sites and it's just link in most cases.. @AliasIO - What do you suggest? Here is another example site - https://www.blueskyelearn.com/

adityapandey1998: Example - https://www.essentialaccessibility.com/carilionclinic/?utm_source=carilionhomepage&utm_medium=iconlarge&utm_term=eachannelpage&utm_content=header&utm_campaign=carilionclinic

adityapandey1998:
@AliasIO @rockeynebhwani I have made changes to remove .*
Please check

src/apps.json
AliasIO: I'm not sure if a link is enough to say the website is using the technology but perhaps including the icon is specific enough:

`"<a href=\"[^\"]+essentialaccessibility\\.com[^>]+><img src=\"[^\"]+eSSENTIAL-ACCESSIBILITY-ICON"`

src/apps.json
```diff
@@ -1178,6 +1195,18 @@
       "icon": "Atlassian Statuspage.svg",
       "website": "https://www.statuspage.io/"
     },
+    "AudioEye": {
+      "cats": [
+        68
+      ],
+      "html": [
+        "<iframe[^>]*audioeye\\.com/frame/cookieStorage",
+        "ws\\.audioeye\\.com/ae\\.js"
```
AliasIO, line 1204: This doesn't match any HTML tags and can probably be removed as it's covered by `script`.

src/apps.json
adityapandey1998: This image isn't present in other sites though.
Sites can name the images differently

src/apps.json
adityapandey1998: Got it

adityapandey1998: There were a couple of sites that wasn't being covered in script for some reason. That's why I included the iframe component.
Will remove the JS part

adityapandey1998:
@AliasIO @rockeynebhwani Have made required changes
Do review this PR

src/apps.json
adityapandey1998: I have removed this for now.

rockeynebhwani: @AliasIO - We couldn't find reliable ways to track 'Essential Accessibility as each site had different names.. So, we have removed this for now and would like other changes to be merged if you are happy

AliasIO:
Thanks!

PR #3263, opened 2020-07-30
add Cloudinary in CDN category

phucnh-nfq:


phucnh-nfq:
Some website are using this
https://ubuntu.com
https://www.crunchbase.com
https://cloudinary.com
https://www.channelnewsasia.com
https://www.pgatour.com
https://www.konga.com
https://www.smashingmagazine.com
https://www.c-and-a.com/eu/en/shop
https://www.1mg.com
https://www.powtoon.com
https://splice.com
https://www.fiu.edu

AliasIO:
This will identify the CDN on any website that embeds an image on the domain. Is there a more accurate way to detect it, .e.g a response header or cookie?

ThomasLohner:
Cloudinary images are usually just used in img-tag. The page html is not served from cloudinary, this means we won't see any headers or cookies. I totally see what you mean but on the other hand i don't see a reason why someone would use images from their domain without using the service ;-)

AliasIO:
Sorry for neglecting this PR, I included this manually as master has diverged a lot. Thank you!

phucnh-nfq:
Thank you for accepted.

Issue #3246, opened 2020-07-24
Improve coverage of Tag Managers


rockeynebhwani:
Add following Tag Managers - 

Technology | Website | Pattern | Example site
------------ | ------------- | ------------- | -------------
IBM DDX|https://www.ibm.com/support/knowledgecenter/en/SSPG9M/DDX/DDX/ddx_intro_kc.html|?|?

These Tag Managers are already present in apps.json but are under other categories as a vendor provides more than one solution. 

Technology | Website | Pattern | Example site
------------ | ------------- | ------------- | -------------
Matomo / Piwik Pro | https://matomo.org/docs/tag-manager|JS function window.MatomoTagManager|https://www.georgeboom.com/
Tealium|https://tealium.com/products/tealium-iq-tag-management-system/|?|?
Signal|https://signal.co/customer-intelligence-solutions/tag-management/|No additional pattern required. Already in apps.json|https://direct.asda.com

nurbek91:
@AliasIO Hey mate .Signal, Tealium , Matomo / Piwik Pro  - Looks like it's already implemented in tech.json 
#3504 newly added . Otherwise, there are not many of them.

PR #3272, opened 2020-08-01
Add AEP Launch and refine detection of DTM

janexner:
Added detection of Adobe Experience Platform Launch, the successor to DTM, and refined detection of both, DTM, and Launch


janexner:
This could be seen as part of #3246

AliasIO:
Thanks!

PR #3291, opened 2020-08-13
Update Pantheon header detection

kyletaylored:
- Added "x-styx-req-id" as additional header detection for Pantheon sites.


AliasIO:
Thanks!

PR #3306, opened 2020-08-23
Add wordpress version detection from scripts

epicfaace:
Wordpress will often have its version exposed through a script tag such as the following:
 
```
<script type='text/javascript' src='https://.../wp-includes/js/wp-embed.min.js?ver=5.4.2'></script>
```


AliasIO:
Thanks!

PR #3314, opened 2020-08-28
Recognize NitroPack as a caching solution

raxbg:


AliasIO:
Thanks for the PR! Would you be able to add the new `description` field as well?

https://www.wappalyzer.com/docs/dev/specification/#optional-properties

raxbg:
Yep, that makes sense :smiley: 

Should be good now

AliasIO:
Thanks!

PR #3326, opened 2020-09-08
Fixed regex in slugify method.

AlbinAxtelius:
Fixed an issue with the slugify method. 

Currently if you input `"-some string-"` you get `"some-string-"` when you expect to get `"some-string"`.

This PR fixes this issue by adding a global flag on the last replace regex in the slugify method.


AliasIO:
Thanks!

PR #3330, opened 2020-09-08
Add Mixpanel version detection

### LINKED ISSUE #3329: Add Mixpanel version
Adding version to where it can get shown.

Example site:
- https://www.zenhub.com

honjes:
- fixes #3329

adding a new script detection for urls like `cdn.mxpnl.com/libs/mixpanel-'version'.min.js`


AliasIO:
Thanks!

PR #3336, opened 2020-09-10
Add more Font Awesome detection

### LINKED ISSUE #3335: Wrong version Detection Font Awesome
If a sites load fontawesome over the new cdn `kit.fontawesome.com` its detecting the wrong version, because of the simular url.

Example Url:
https://kit.fontawesome.com/192489156d.js

Example sites:
https://stockcharts.com/
https://fontawesome.com/

honjes:
Adding detection for the new kit.fontawesome link and script.
Removing version detection for the script cause of false positives.

- fixes #3335.


AliasIO:
Thanks!

PR #3343, opened 2020-09-16
Add support for other GOV.UK products and refine detection of Frontend

Nooshu:
This PR does the following:

- Refines the detection of GOV.UK Frontend as we were getting lots of false positives off copy / pasted cookie banners
- Adds the detection of GOV.UK Frontend's version number if a service team is using the `dist` version of the CSS or JS
- Adds detection of three other GOV.UK Products: GOV.UK Template, GOV.UK Elements, GOV.UK Toolkit
- All products are now use the 'UI Frameworks' category, rather than doubling up in 'Miscellaneous'

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -4808,17 +4808,69 @@
       "icon": "G-WAN.png",
       "website": "http://gwan.com"
     },
-    "GOV.UK Frontend": {
+    "GOV.UK Elements": {
       "cats": [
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks! I removed the references to `window.` as they're redundant.

Nooshu:
Fantastic. Thank you!

PR #3344, opened 2020-09-17
Adding regex to identify Yottaa CDN

ankur-moovweb:


AliasIO:
Hi Ankur, can you please provide an SVG icon and description as well?

Issue #3345, opened 2020-09-17
[NPM Driver]: Timers never get closed unless `Promise.race()` is timed out


BR0kEN-:
**Platform**
Which platform does the bug apply to?
* NPM

**Describe the bug**
The `Promise.race()` is used incorrectly with `setTimeout()` so even if the `Site.analyze()` completes in 1 second the Node process will exit only after all timers are closed (see `maxWait`).

All `Promise.race()` in the `src/drivers/npm/driver.js` looks this way:

```
await Promise.race([
  this.timeout(),
  <PROMISE>,
])
```

The `timeout()` method uses `setTimeout()` internally and all set timers never get cleared if `<PROMISE>` wins the race therefore the Node process lasts at least the amount of seconds configured via `maxWait` and not less.

**To Reproduce**
- Create the `test-timers.js` file with the following contents:

  ```javascript
  const timeout = 5000;
  
  async function examplePromise() {
    return new Promise((resolve) => setTimeout(resolve, 1000));
  }
  
  async function raceIncorrect() {
    return Promise.race([
      new Promise((resolve, reject) => setTimeout(reject, timeout, new Error('Timed out.'))),
      examplePromise(),
    ]);
  }
  
  async function raceCorrect() {
    let timerId = null;
  
    return Promise.race([
      new Promise((resolve, reject) => {
        timerId = setTimeout(() => {
          clearTimeout(timerId);
          reject(new Error('Timed out.'));
        }, timeout);
      }),
      examplePromise().then((value) => {
        clearTimeout(timerId);
        return value;
      }),
    ]);
  }
  
  raceIncorrect();
  ```
- Run it with `time node test-timers.js` and see the process completes in 5+ seconds.
- Replace call to `raceIncorrect()` by `raceCorrect()` and see the process completes for 1 second and some milliseconds.

**Expected behavior**
The timers should not prevent the Node process from exiting.

BR0kEN-:
@AliasIO, please take a look at the https://github.com/AliasIO/wappalyzer/pull/3346 when you have some spare time.

AliasIO:
Good call, thanks for the PR!

PR #3346, opened 2020-09-17
#3345: Close a timer after resolving/erroring the `Promise.race()`

BR0kEN-:
https://github.com/AliasIO/wappalyzer/issues/3345


AliasIO:
That looks much better indeed, thank you!

PR #3347, opened 2020-09-18
Update mobX and Symfony regexp.

RomaLash:
Update mobX regexp. Now it is possible to get version.

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -12580,7 +12580,7 @@
         "sf_redirect": ""
       },
       "description": "Symfony is a PHP web application framework and a set of reusable PHP components/libraries.",
-      "html": "(?:<div class=\"sf-toolbar[^>]+?>[^]+<span class=\"sf-toolbar-value\">([\\d.])+|<div id=\"sfwdt[^\"]+\" class=\"[^\"]*sf-toolbar)\\;version:\\1",
+      "html": "(?:<div class=\"sf-toolbar[^>]+?>.*?<span class=\"sf-toolbar-value\">|<div id=\"sfwdt[^\"]+\" class=\"[^\"]*sf-toolbar)",
       "icon": "Symfony.svg",
       "implies": "PHP",
       "js": {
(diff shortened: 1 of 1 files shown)


RomaLash:
Old regexp for sypfony was weak. Thats why there was so much false positive matches.

For example:

`div class="sf-toolbar-block sf-toolbar-block-config sf-toolbar-status-normal sf-toolbar-block-right" title="">
...
 <span class="sf-toolbar-value">4.4.2</span>
...
</div>`

This piece of HTML contains version for Symfony (thats what regexp looking for), but :

`
 <div class="sf-toolbar-block sf-toolbar-block-cache sf-toolbar-status-normal ">
...
   <span class="sf-toolbar-value">2117</span>
...
</div>`

this code contains smth related with cache and etc. But old regexp matched this too. In result we have Symfony version 2117.
So we have to remove part of regexp, which contains version matching.

src/technologies.json
```diff
@@ -8249,7 +8249,7 @@
       "icon": "MobX.svg",
       "js": {
         "__mobxGlobal": "",
-        "__mobxGlobals": "",
+        "__mobxGlobals.version.toString()": "([\\d.]+)\\;version:\\1",
```
AliasIO, line 8252: This should probably just b`__mobxGlobals.version` as `toString()` won't work. Functions are not executed on the page.

RomaLash:
I think we have a little missunderstanding.
U said that "Functions are not executed on the page"

But what about this https://github.com/AliasIO/wappalyzer/blob/5892c933ef347526adfc6d5767b77644b5022c8a/src/drivers/npm/driver.js#L487

Maybe my Q is so stupid, cos i dont know js well. 
I tried to port this project to python.

AliasIO:
The extension does run JavaScript on web pages but the `js` pattern don't get executed. So `__mobxGlobals.version` works but `__mobxGlobals.version()` doesn't.

PR #3352, opened 2020-09-19
Update category ID for Tealium

davidgnavas:
The most appropriate category for Tealium is Tag managers because although its technology helps to integrate marketing solutions, it does not only integrate advertiser services but also analytics, cookie compliance, data management... In short, everything a tag manager does.


AliasIO:
Thanks!

PR #3363, opened 2020-09-20
Add Flowvy by Nethit Systems

villeristi:


src/technologies.json
```diff
@@ -4614,6 +4614,17 @@
...
       "website": "https://flickity.metafizzy.co/"
     },
+    "Flowvy": {
+      "cats": [
+        6
+      ],
+      "cookies": {
+        "NHsect": ""
+      },
+      "html": "<a href='https://www.flowvy.com' target='_blank'>",
```
AliasIO, line 4624: Can this be improved? A link to this website doesn't mean the technology is in use.

villeristi: Actually it does as the engine is non-modifiable and the link is hardcoded into the templates.

AliasIO: But any website could link to flowvy.com, which doesn't mean the technology is in use.

PR #3365, opened 2020-09-20
Add Classy detection

oowl:
[Classy](https://www.classy.org/) is used by more than 4,000 US nonprofit organizations to build their white-label fundraising pages.


AliasIO:
Thanks!

PR #3381, opened 2020-09-25
Add Linkedin CDN Detection

### LINKED ISSUE #3380: Add LinkedIn CDN
- Script: snap.licdn.com/
- Example: https://www.truework.com/

honjes:
- fixes #3380

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -7302,6 +7302,14 @@
       "scripts": "//platform\\.linkedin\\.com/in\\.js",
       "website": "http://linkedin.com"
     },
+    "Linkedin CDN": {
+      "cats": [
+        31
+      ],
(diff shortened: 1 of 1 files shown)


AliasIO:
Is there a way to detect this CDN via a header, and would it be used by any site other than LinkedIn? I'd like to avoid detecting CDNs on websites that merely load a script or image from a CDN URL.

honjes:
I could not find a header.

The CDN is used to load the Linkedin Analytics `https://snap.licdn.com/li.lms-analytics/insight.min.js` example on `https://www.truework.com/`.

If you like too avoid this kind of detectionm then this CDN should not be detection.

AliasIO:
I'll update these to detected Linkedin Insight instead, thank you. :)

PR #3389, opened 2020-09-30
improved doc

barinder7:


AliasIO:
Thank you but we use "ecommerce" without a dash throughout. :)

PR #3402, opened 2020-10-01
Improving gmaps detection

honjes:
Examplesite: 
- www.doordash.com/en-US

Detection Script:
- maps.google.com/maps/api/js


AliasIO:
Thanks!

PR #3420, opened 2020-10-07
Update wrong header's props in Commerce.js

jaepass:
- Updated wrong headers
With many apologies for this update again! 🙏

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -3101,13 +3101,13 @@
       ],
       "description": "Commerce.js is an API-first eCommerce platform for developers & businesses.",
       "headers": {
-         "X-Support":  "support@commercejs\\.com",
+         "X-Support": "support@commercejs\\.com",
          "Chec-Version": ".*"
        },
       "icon": "commercejs.png",
       "scripts": [
-        "^https?:/cdn\\.chec\\.io/v\\d+/commerce\\.js",
-        "chec/commerce.jscommerce\\.js"
+        "https?:/cdn\\.chec\\.io/v(\\d+)/commerce\\.js\\;version=\\1",
+        "chec/commerce\\.js"
       ],
       "url": "^https?//.+\\.spaces.chec\\.io",
       "website": "https://www.commercejs.com"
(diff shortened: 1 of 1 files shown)


src/technologies.json
```diff
@@ -3101,11 +3101,12 @@
       ],
       "description": "Commerce.js is an API-first eCommerce platform for developers & businesses.",
       "headers": {
-        "X-Powered-By": "Commerce.js"
-      },
+         "X-Support":  "support@commercejs\\.com",
+         "Chec-Version": ".*"
```
AliasIO, line 3105: Can this be used to capture the version number? E.g. `"Chec-Version": "(\\d+)"\\;version=\\1`

src/technologies.json
```diff
@@ -3101,11 +3101,12 @@
...
+       },
       "icon": "commercejs.png",
       "scripts": [
-        "^https?:/cdn\\.chec\\.io/v2/commerce\\.js/",
+        "^https?:/cdn\\.chec\\.io/v\\d+/commerce\\.js",
```
AliasIO, line 3109: Or this? E.g. `"^https?:/cdn\\.chec\\.io/v(\\d+)/commerce\\.js\\;version=\\1`

src/technologies.json
jaepass: hey yes thanks for the suggestion, that might be ideal. But do do have version 1 and 2 in the wild

ScopeyNZ: The `Chec-Version` header is actually a date version that our API provides, not our SDK version. There's two versions in play here, I'm not sure if we can differentiate them in wappalyzer. I think the SDK version is more appropriate.

src/technologies.json
jaepass: thanks a ton @AliasIO for the suggestion, i've just pushed it up! 🙏🙏🙏

jaepass:
Let me know if all looks good thanks again!

AliasIO:
Thanks!

PR #3421, opened 2020-10-07
Fix typo in GOV.UK Frontend version number

Nooshu:
Version number detected but only returns `a` because of typo copied from example in docs.

Apologies for this minor update, I'm frustrated I didn't spot it sooner.


AliasIO:
Thanks!

PR #3426, opened 2020-10-14
Zone.js implies Angular

jxxe:
Google Fonts uses Angular, but Angular is not detected. However, Zone.js is, which is a part of Angular.


AliasIO:
Thanks!

PR #3440, opened 2020-10-24
Support Telerik UI, improve Sitefinity, add tests

epicfaace:
- Support version detection for Telerik UI for ASP.NET
- Improve Sitefinity version detection -- meta tags with "Sitefinity 11.0.6700.0 PE" will now be detected as having a version of "110.0.6700.0" instead of "11.0.6700.0 PE"
- Add tests. I've added a declarative way to add tests with the `examples` keyword, so that we can make sure that new regexes actually work on sample test cases and that different version detectors won't interfere with each other. I've used some of the logic from https://github.com/Kikobeats/simple-wappalyzer to actually do this from custom HTML in a fast way. We can run these tests on PRs to ensure that things don't break. @AliasIO let me know what you think and if you like this approach, I could add documentation for it.


AliasIO:
Thanks, I appreciate your work on this. I think creating comprehensive test cases that also cover `dom`, `js` and `dns` will require a fair bit of maintenance as we're essentially duplicating all driver functionality. `js` and `dom` are the most important ones to cover as `html` is now deprecated. Because of this I won't be merging this, although I do like the idea. Do you want to submit a separate PR for your Telerik UI and Sitefinity changes?

epicfaace:
That's interesting, why is `html` deprecated? Detecting comments (`<!-- ... -->`) (which is helpful for Telerik UI version detection) appears to only be possible with `html` and not with `dom`.

AliasIO:
The `dom` method is more efficient and covers almost everything `html` does. You're right that HTML is currently the only way to inspect comments.

PR #3456, opened 2020-10-28
Improve Confluence vs Jira detection

theconcierge2020:
Hey guys,

using wappalyzer recently I discovered, that a lot of times websites get marked as Atlassian Confluence AND Jira. However, only the Jira technology is capable of finding a version and the Confluence was not.
This is happening, because both applications have the "ajs-version-number" (which is stated within the Atlassian Jira app rule). As well, the provided html-regex ("Powered by...") is not working - at least on pages I encountered. Tested it as well manually.

So my suggestion for lower false-positives would be:

- Remove "ajs-version-number" from JIRA - because both apps have it.
- Jira pages have one attribute within a meta-tag and the body-tag: "data-version", which we can take advantage of -> therefore changed the html-regex
- Confluence pages have in addition to "Powered by ..." the "Printed by Atlassian Confluence [_version_]" within the footer section. However, depending on the language the only useable part is "Atlassian Confluence [_version_]" ("Printed by" gets translated). I guess the "Powered by..." regex could be revised as well, but in my opinion it does make the regex simpler for the same result - this is why I suggest using it.

It is a proposal and I wanted feedback from the community about this. 
If you have any input, let me know and I will work on it. 

Greetings

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -1640,7 +1640,7 @@
       "headers": {
         "X-Confluence-Request-Time": ""
       },
-      "html": "Powered by <a href=[^>]+atlassian\\.com/software/confluence(?:[^>]+>Atlassian Confluence</a> ([\\d.]+))?\\;version:\\1",
+      "html": "Atlassian Confluence ([\\d.]+)\\;version:\\1",
       "icon": "Atlassian Confluence.svg",
       "implies": "Java",
       "meta": {
@@ -1665,14 +1665,13 @@
         13
       ],
       "cpe": "cpe:/a:atlassian:jira",
-      "html": "Powered by\\s+<a href=[^>]+atlassian\\.com/(?:software/jira|jira-bug-tracking/)[^>]+>Atlassian\\s+JIRA(?:[^v]*v(?:ersion: )?(\\d+\\.\\d+(?:\\.\\d+)?))?\\;version:\\1",
+      "html": "data-version=\"([\\d.]+)\\;version:\\1",
       "icon": "Atlassian Jira.svg",
       "implies": "Java",
       "js": {
         "jira": ""
       },
       "meta": {
-        "ajs-version-number": "^(.+)$\\;version:\\1",
         "application-name": "JIRA"
       },
       "website": "http://www.atlassian.com/software/jira/overview/"
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -1640,7 +1640,7 @@
       "headers": {
         "X-Confluence-Request-Time": ""
       },
-      "html": "Atlassian Confluence ([\\d.]+)\\;version:\\1",
+      "html": "<li class=\"print-only\">.+ Atlassian Confluence ([\\d.]+)</li>\\;version:\\1",
       "icon": "Atlassian Confluence.svg",
       "implies": "Java",
       "meta": {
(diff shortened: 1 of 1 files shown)


src/technologies.json
```diff
@@ -1640,7 +1640,7 @@
       "headers": {
         "X-Confluence-Request-Time": ""
       },
-      "html": "Powered by <a href=[^>]+atlassian\\.com/software/confluence(?:[^>]+>Atlassian Confluence</a> ([\\d.]+))?\\;version:\\1",
+      "html": "Atlassian Confluence ([\\d.]+)\\;version:\\1",
```
AliasIO, line 1643: HTML patterns must include a `<` or `>` to avoid matching plain text. Can you include some of the surrounding HTML?

src/technologies.json
```diff
@@ -1665,14 +1665,13 @@
         13
       ],
       "cpe": "cpe:/a:atlassian:jira",
-      "html": "Powered by\\s+<a href=[^>]+atlassian\\.com/(?:software/jira|jira-bug-tracking/)[^>]+>Atlassian\\s+JIRA(?:[^v]*v(?:ersion: )?(\\d+\\.\\d+(?:\\.\\d+)?))?\\;version:\\1",
+      "html": "data-version=\"([\\d.]+)\\;version:\\1",
```
AliasIO, line 1668: Same here, please include the HTML tag as well. Thanks!

src/technologies.json
theconcierge2020: Done - please review :)

src/technologies.json
theconcierge2020: Done - please review :) 

Also added one meta-tag regex.

src/technologies.json
```diff
@@ -1665,15 +1665,15 @@
         13
       ],
       "cpe": "cpe:/a:atlassian:jira",
-      "html": "Powered by\\s+<a href=[^>]+atlassian\\.com/(?:software/jira|jira-bug-tracking/)[^>]+>Atlassian\\s+JIRA(?:[^v]*v(?:ersion: )?(\\d+\\.\\d+(?:\\.\\d+)?))?\\;version:\\1",
+      "html": "<body.+data-version=\"([\d.]+)\">\\;version:\\1",
```
AliasIO, line 1668: `<body.+` should be `<body[^>]+` to keep the match inside the tag.

src/technologies.json
```diff
@@ -1640,7 +1640,7 @@
       "headers": {
         "X-Confluence-Request-Time": ""
       },
-      "html": "Powered by <a href=[^>]+atlassian\\.com/software/confluence(?:[^>]+>Atlassian Confluence</a> ([\\d.]+))?\\;version:\\1",
+      "html": "<li class=\"print-only\">.+ Atlassian Confluence ([\\d.]+)</li>\\;version:\\1",
```
AliasIO, line 1643: Try to avoid `.+` as it can match very large strings. If the string inside `<li>` isn't known you can use something like`.{0,50}` to limit the length of the string that will be matched.

AliasIO:
Thanks!

theconcierge2020:
So sorry for the extreme late response! Not the nice way - my biggest apologies. 

I really like this project and will try to contribute in the future more. 
If you need anything - let me know!

Best to you

PR #3457, opened 2020-10-28
Add Bookingkit/Appointment, Booking widget

nurbek91:
Website:
https://bookingkit.net/
Example websites:
http://die-jagd-hat-begonnen.de/
http://www.rausundrauf.com/en
http://lanciola.it/


src/technologies.json
```diff
@@ -2175,6 +2175,24 @@
...
+        72
+      ],
+      "description": "Bookingkit is an online booking management solution. Bookingkit helps its users generate PDF invoices, manage day-to-day scheduling operations, and automatically sync availabilities in real time.",
+      "icon": "Bookingkit.svg",
+      "js": {
+        "BookingKitApp": ""
+      },
+      "html": [
+        "<style type=\"text/css\">\\#bookingKitContainer",
+        "bookingkit\\.css",
```
AliasIO, line 2191: HTML patterns should include a `<` or `>` so to avoid matching plain text. Does this appear inside a `<link>` tag?

src/technologies.json
```diff
@@ -2175,6 +2175,24 @@
...
+        61,
+        72
+      ],
+      "description": "Bookingkit is an online booking management solution. Bookingkit helps its users generate PDF invoices, manage day-to-day scheduling operations, and automatically sync availabilities in real time.",
+      "icon": "Bookingkit.svg",
+      "js": {
+        "BookingKitApp": ""
+      },
+      "html": [
+        "<style type=\"text/css\">\\#bookingKitContainer",
```
AliasIO, line 2190: Are the `\\` needed?

You can probably also use `"css": "#bookingKitContainer"` here and maybe `"dom": { "#bookingKitContainer": { "attributes" { "id": "" } }`

PR #3459, opened 2020-10-28
Add Booksy/Appointment, Booking widget

nurbek91:
Website:
https://booksy.com/
Example Websites:
http://intimatewhisperhairsalon.com/
http://www.salonindividualist.com/

@AliasIO  hey mate . 
1 ) Do I need to add this option: **https://javitorrentebarbers.es/** . 
There's just a link transition. The widget itself on the site itself is not implemented
2) I did not understand how  can take the version. Or is it not possible to take it? 
**https://booksy.com/widget/code.js?id=91945&country=us&lang=en**


AliasIO:
Avoid matching links if possible, in this case it doesn't need to be included.

You can access the `version` variable from outside the function so don't worry about it.

PR #3464, opened 2020-10-29
Update JET NEO detection

luiz-gustavo:
Website:
https://www.jetecommerce.com.br/
Example Websites:
https://frameworkneo-v2.plataformaneo.com.br/
https://www.bynv.com.br/
https://brooksdonna.com.br/

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -9093,11 +9093,20 @@
       "cats": [
         6
       ],
+      "cookies": {
+        "SessionIdCookie": ""
+      },
+      "description": "NEO is an ecommerce software that manages multiple online stores.",
       "headers": {
         "powered": "jet-neo"
       },
-      "icon": "Plataforma NEO.svg",
-      "website": "http://www.jetecommerce.com.br/"
+      "html": [
+        "<a [^>]*href=\"[^/]*//[^/]*plataformaneo\\.com",
+        "<img [^>]*src=\"[^/]*//[^/]*plataformaneo\\.com"
+      ],
+      "icon": "plataforma_neo.svg",
+      "url": "(^|^[^:]+:\/\/|[^\\.]+\\.)plataformaneo\\.com",
+      "website": "https://www.jetecommerce.com.br"
     },
     "NVD3": {
       "cats": [
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -9100,10 +9100,11 @@
       "headers": {
         "powered": "jet-neo"
       },
-      "html": [
-        "<a [^>]*href=\"[^/]*//[^/]*plataformaneo\\.com",
-        "<img [^>]*src=\"[^/]*//[^/]*plataformaneo\\.com"
-      ],
+      "dom": {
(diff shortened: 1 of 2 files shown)


src/technologies.json
```diff
@@ -9093,11 +9093,20 @@
...
+      "description": "NEO is an ecommerce software that manages multiple online stores.",
       "headers": {
         "powered": "jet-neo"
       },
-      "icon": "Plataforma NEO.svg",
-      "website": "http://www.jetecommerce.com.br/"
+      "html": [
+        "<a [^>]*href=\"[^/]*//[^/]*plataformaneo\\.com",
+        "<img [^>]*src=\"[^/]*//[^/]*plataformaneo\\.com"
+      ],
```
AliasIO, line 9106: Links and images are usually not enough to confidently say the website is using the technology. Can this be made more specific?

src/technologies.json
```diff
@@ -9093,11 +9093,20 @@
...
+      "icon": "plataforma_neo.svg",
+      "url": "(^|^[^:]+:\/\/|[^\\.]+\\.)plataformaneo\\.com",
```
AliasIO, line 9108: Could this made simpler, e.g. "\\.plataformaneo\\.com"?

src/technologies.json
luiz-gustavo: the code is open front available to agencies and only these items guarantee the identification of the platform :(

src/technologies.json
luiz-gustavo: Can be.
I will change.
Tks!

src/technologies.json
AliasIO: What about the link text? Does the link say something like 'powered by JET NEO'?

luiz-gustavo: The link are images of products to help in the identification because the plugin stopped identifying the platform.
I inserted it because the plugin stopped identifying the platform as in the past using the header.
Take this site as an example.
Powered by "jet-neo" has been present at json since 2017 but still does not identify.
https://www.bynv.com.br/
https://www.lembrancachique.com.br/

AliasIO: I found some HTML on the checkout pages that we can use. This should work:

```
"dom": {
  "[data-jet-validate], #svr[value="NEOWEBV4"]": {
    "text": ""
  }
}
```

src/technologies.json
```diff
@@ -9093,11 +9093,20 @@
       "cats": [
         6
       ],
+      "cookies": {
+        "SessionIdCookie": ""
```
Johannes-Andersen, line 9097: Is this really a safe way of checking for NEO without false positives? I would imagine `SessionIdCookie` is not unique to this site based on how generic the name is.

AliasIO: Agree, it's pretty generic and it should probably be left out.

src/technologies.json
luiz-gustavo: Changed!

AliasIO:
Thanks!

PR #3471, opened 2020-11-02
Add Formitable/Booking,Payment widget

nurbek91:
Website:
https://formitable.com/
Example websites:
https://www.loroblonyo.nl/
http://abrahamkef.nl/
https://alohabeach.nl/
https://akropolis-geleen.nl/
I have a little problem here, I can't take the version, can u show me how to do it correctly
https://www.taikocuisine.com/en/home-2/
 https://www.taikocuisine.com/wp-content/themes/taiko-v2/js/formitable.js?ver=4.9.15


AliasIO:
This should work: `"formitable\\.js(?:\\?ver=([\\d.]+))?\\;version:\\1"`

nurbek91:
> This should work: `"formitable\\.js(?:\\?ver=([\\d.]+))?\\;version:\\1"`

image
Could you tell me why this sign is worth?
Made to work for cases when the link does not contain a version

image

I'm a little confused and don't understand when I can specify just like that *  after .js  and when I need to write a full mask
Sorry in advance for the silly questions. Regular expressions can be trained endlessly ;)
By the way checked  and committed your version regex

AliasIO:
`*` Means zero or more. So `.js*` matches `.j`, `.js` and `.jsss`, which is not what you want. You could do `(.js)*`but that also matches `.js.js`.
 
The `?` makes something optional, so `\.js(\?ver=)?` matches `.js` and `.js?ver=`.

PR #3472, opened 2020-11-02
Add wpBakery page builder

nurbek91:
Website:
https://wpbakery.com/
Example websites:
https://factcheck.kz/
https://www.wltz.com/
https://liter.kz/

Also there are a lot of sites here where the meta is not defined , but the html document is attached. I didn't take them. But if you say that we need to take them and **implies** reduces loads, I'll turn them on too 
https://marketplace.walmart.com/
image


AliasIO:
Thanks, I think this is fine as it is.

Issue #3370, opened 2020-09-24
LogRocket


AndyDaSilva52:
[LogRocket WebSite](https://logrocket.com/)

PR #3492, opened 2020-11-08
feat(LogRocket): add LogRocket (#3370)

Johannes-Andersen:
Add LogRocket (https://logrocket.com/) as requested by AndyDaSilva52 in #3370

Customer list can be found at https://logrocket.com/customers/ Should meet criteria of widely used technologies﻿

Most websites only uses LogRocket on signed in users, however by checking https://trends.builtwith.com/websitelist/LogRocket https://media.alfred.com/ use logrocket for every visitor, so you can use that for testing 

Screenshot 2020-11-08 at 00 35 22


AliasIO:
Thanks!

PR #3494, opened 2020-11-08
fix(Rakuten): duplicate Rakuten entery

Johannes-Andersen:
Fix following error:

```
$ yarn run validate
yarn run v1.22.10
$ yarn run lint && jsonlint -qV ./schema.json ./src/technologies.json && node ./bin/validate.js
$ eslint --fix src/**/*.js
File: src/technologies.json
Parse error on line 11266, column 16:
...  },    "Rakuten": {      "cats": [  ...
----------------------^
/technologies/Rakuten should NOT have additional properties; see #/properties/technologies/additionalProperties/additionalProperties
```

Rakuten had a duplicate entry﻿

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -11779,23 +11779,6 @@
       },
       "website": "https://www.webgains.com/"
     },
-    "Rakuten": {
-      "cats": [
-        71,
-        61
-      ],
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

Issue #3350, opened 2020-09-19
Add AOS detection


honjes:
sites:
- https://michalsnik.github.io/aos/
- https://github.com/michalsnik/aos

example sites:
- https://michalsnik.github.io/aos/
- https://www.georgeboom.com/

can be detected with the `/aos@next/dist/aos.js` where 'next' could be a aversion number

PR #3495, opened 2020-11-08
feat(AOS): add AOS (#3350)

Johannes-Andersen:
Add AOS detection as requested by honjes in #3350 

AOS has 14k stars, https://github.com/michalsnik/aos, so should meet notability requirements 

Works on all of the sites here https://trends.builtwith.com/websitelist/AOS 

Added the confidence for "js" as AOS is kinda short and may have conflicts with minified code, but should be rather safe if all 3 of the methods are there﻿

Screenshot 2020-11-08 at 03 40 09


AliasIO:
Thanks!

PR #3500, opened 2020-11-08
fix(technologies): resolve duplicate entries

Johannes-Andersen:
It seems like someone is creating duplicate entries or not changing the object key when adding new technologies

This pull request will remove the duplicate entries, and rename the object key where it was already existing, but the values did not seem to be a duplicate.

Is it worth thinking about adding automated tests when creating a commit to prevent this from happening?﻿


AliasIO:
Thanks, it's probably me fixing up merge conflicts by hand. I did just fix some of these.

PR #3501, opened 2020-11-09
feat(technologies): add fraud provider device JS

Johannes-Andersen:
Fraud providers offer services where they have merchants include tracking scripts. These are some of the major players in the e-commerce fraud space﻿


AliasIO:
Thanks! The icons need to be square but I'll edit them.

PR #3514, opened 2020-11-12
Axios improve + move from misc to JavaScript lib

Johannes-Andersen:
Related to issue #3509, as axios is already being detected, I just added a new check to help improve a lot of missed detections

Also moved to a better category

Axios has no logo, so did not add a logo﻿


AliasIO:
👍

PR #3517, opened 2020-11-12
Add CoreMedia CMS detection

greenoid:
CoreMedia is a Hamburg, Germany based software company which offers a professional web content management system for big companies. In the past simply named CoreMedia CMS is now know as "CoreMedia Content Cloud". See their website
https://www.coremedia.com/

These are their references:
https://www.coremedia.com/en/customers/iconic-brands

They are here on GitHub too:
https://github.com/CoreMedia

Since the CoreMedia CMS is front-end agnostic (which is a good thing), it is rather difficult to detect. I detect it by the URLs of there resources because they are rather unique.

I cannot detect all the reference sites, because some of them use the CoreMedia as a headless background CMS to manage their digital assets, but the front is running a shop system. Their CMS comes with a range of integration adapters for other systems like shops, so this is to be expected.

I managed to detect it in these rather big websites:
https://www.o2online.de
https://www.deutschebahn.com/
https://www.dbschenker.com/de-de/
https://www.telekom.com/
https://www.hamburg.de/
https://www.tk.de/
https://www.freenet.de/index.html
https://www.continental.com/en
https://www.tdk-electronics.tdk.com/en
https://www.finnair.com/de-de
https://tommy.com/

I hope the code is fine and it is merged. I used wappalyzer for some time with pleasure and would be happy to contribute.

Thanks, greenoid.


src/technologies.json
```diff
@@ -3476,6 +3476,25 @@
...
+      "cats": [
+        1
+      ],
+      "cookies": {
+        "CM_SESSIONID": "",
+        "CMSESSIONID": ""
+      },
+      "description": "CoreMedia CMS is a web content management system with many of integration connectors for ecommerce shop systems and others.",
+      "html": [
+        "/blueprint/",
```
Johannes-Andersen, line 3489: blueprint is an extremely common word. How can you guarantee a website is using CoreMedia by them just having that path in their HTML?

src/technologies.json
```diff
@@ -3476,6 +3476,25 @@
...
+        "/resources/cae/.*\"",
+        "/resource/(image|css|js|blob|crblob)/[\\d]+/[0-9a-z]+/[^\"]*\\.(jpg|jpeg|png|gif|ico|css|js|json)\"",
+        "/(image|blob|crblob)/[0-9]+/[\\D]*[\\d]+x\\d[^\"]*\\.(jpg|jpeg|png|gif)\""
```
Johannes-Andersen, line 3492: Screenshot 2020-11-12 at 20 31 42
When looking into the regex, it is going to be hella flaky. Like any website may have that format on urls.

Is there no other way of checking JS methods or something better?

src/technologies.json
```diff
@@ -3476,6 +3476,25 @@
...
+    "CoreMedia CMS": {
+      "cats": [
+        1
+      ],
+      "cookies": {
+        "CM_SESSIONID": "",
+        "CMSESSIONID": ""
+      },
+      "description": "CoreMedia CMS is a web content management system with many of integration connectors for ecommerce shop systems and others.",
+      "html": [
```
Johannes-Andersen, line 3488: The HTML selection method has been deprecated as of https://github.com/AliasIO/wappalyzer/pull/3440#issuecomment-717586317

src/technologies.json
Johannes-Andersen: Example of the issue: Screenshot 2020-11-12 at 20 36 38

By going to https://pastebin.com/JLzNb10A this selector will false positive it as finding CoreMedia

greenoid: I can do without "/blueprint/". You may just remove it. It is seldom used. More than often "/blueprint/servlet/" is mapped to "/" on the outside.

src/technologies.json
greenoid: Jep. You are right. So skip this one too. I may come up with a more specific version to catch responsive images, resized by the CMS like in this URL:
/resource/image/83396/16x9/570/321/b39482b0b55a83aaec679c490c9e6cac/sJ/foto2-hsv-und-hansemerkur-verlaengern-partnerschaft.jpg

src/technologies.json
greenoid: I see. So I have to recode it with dom.
As i wrote CoreMedia CMS is frontend framework agnostic, so no JS hints to it.

Issue #3529, opened 2020-11-15
Unchecked runtime.lastError: The message port closed before a response was received.


matthiasm11:
**Platform**
Which platform does the bug apply to?
* WebExtension (Google Chrome)

**Describe the bug**
When visiting a website, f.e. https://www.google.com, the following error is thrown in the browser console:
`Unchecked runtime.lastError: The message port closed before a response was received.`

**To Reproduce**
1. Go to https://www.google.com
2. Open the console, the error is present.

**Additional context**
The error isn't always being thrown. Doing hard refreshes (CMD+R or CTRL+F5), it is appearing +-25% of the time.
I'm not the only person getting this, the review on November 12th on https://chrome.google.com/webstore/detail/wappalyzer/gppongmhjkpfnbhagpmjfkannfbllamg states the same error.
All other extensions are disabled.

matthiasm11:
Similar old (fixed) issues:
- https://stackoverflow.com/questions/43154963/promise-error-the-message-port-closed-before-a-reponse-was-received
- https://github.com/AliasIO/Wappalyzer/issues/1487
- https://github.com/AliasIO/wappalyzer/issues/1489

kamilkozak:
Same issue here. It appeared from version v6.4.14

AliasIO:
Fixed in v6.5.17.

PR #3530, opened 2020-11-15
Fix runtime.lastError

Johannes-Andersen:
Fix to #3529 

In this commit: https://github.com/AliasIO/wappalyzer/commit/3097ddcde8479980f0929bf61b9bc07749433a84#diff-3a09560d3fab721a41189cfa740746e5f2ee1d7562d30fd862f17d4df9f517ec `ads: await getOption('ads', []),` got replaced with `{}`

Since line 463 had `Driver.cache.ads.push(ad)`﻿it is not going to be able to push to an object

Here is the error that caused it to throw an error when visiting websites:
Screenshot 2020-11-15 at 16 13 34

### DIFF
--- a/src/drivers/webextension/js/driver.js
+++ b/src/drivers/webextension/js/driver.js
@@ -54,7 +54,7 @@ const Driver = {
       ),
       tabs: {},
       robots: await getOption('robots', {}),
-      ads: {},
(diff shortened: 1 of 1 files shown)


AliasIO:
Thank you!

PR #3546, opened 2020-11-17
Add detection of Wagtail

thibaudcolas:
This PR adds detection for [Wagtail](http://github.com/wagtail/wagtail), an open-source CMS based on Django. Here are examples of sites built with Wagtail that can now be detected with Wappalyzer:

- <https://www.consumerfinance.gov/>
- <https://www.shakespeare.org.uk/>
- <https://www.great.gov.uk/>
- <https://donate.mozilla.org/>

The detection is based on a regular expression that attempts to match the URLs of images uploaded into the CMS and served by Wagtail, which have a very distinctive format due to a feature called [image renditions](https://docs.wagtail.io/en/stable/topics/images.html?highlight=renditions#generating-image-renditions-in-python). This does mean that this detection only works on pages that contain user-uploaded images (as opposed to static images). Here are a few examples of URLs the regex tries to match:

```
https://www.jazzfestival.nz/media/images/video-fallback-image_1.width-1920.png
/media/images/bevi_3_mob.original.format-png.png
/media/original_images/Social_Distancing_Dashboard_-_image_1_TU_DelftAMS_Institute.png
https://buckup-ff-stories.s3.amazonaws.com/original_images/Lockup_Logo_-_JPEG-1.png
/media/images/UE4LowPolyVehicleTh.2e16d0ba.fill-960x540-c0.jpg
```

The distinctive features are the folders – always either `images` or `original_images`, either placed directly at the root of a site, or from a `/media` subfolder at the root, or a single level up. And then the fact that all of the site’s images are placed in those two folders without any additional sub-folders, with file name suffixes following a set format for all files under `images/` like `.width-1920` or `.fill-960x450-c0` above. I came up with the particular regex used here after trying out different variations on 800'000 homepages from the [Majestic Million](https://majestic.com/reports/majestic-million) list of domains.

---

For Wappalyzer, I initially implemented this with `html` as:

```json
"html": [
    "<[^>]+/media/original_images/[\\w-]+\\.",
    "<[^>]+/media/images/[\\w-.]+\\.(?:(?:fill|max|min)-\\d+x\\d+(?:-c\\d+)?|(?:width|height|scale)-\\d+|original)\\."
],
```

…and then switched to `dom` since the `html` checks seem to be going away. This has the disadvantage of having to repeat the regex multiple times for all the elements/attributes that could contain image URLs, but I imagine it will be more performant nonetheless, since there’s much less text to match against. I added a very basic attribute selector where possible to make it less likely Wappalyzer has to do the full regex matching.

Here are test pages you can use to try out the different selectors:

| Selector                    | Test page                                                          |
| --------------------------- | ------------------------------------------------------------------ |
| `[style*='images/']`        | <https://b-commerces.fr/>                                          |
| `img[src*='images/']`       | <https://blog.youtube/>                                            |
| `meta[content*='images/']`  | <https://www.yourschoolgames.com/how-it-works/what-school-games/>  |
| `video[poster*='images/']`  | <https://bcw-global.com/>                                          |
| `img[srcset*='images/']`    | Disable all other selectors first - <https://hermitage.nl/nl>      |
| `source[srcset*='images/']` | Disable all other selectors first - <http://www.viktor-rolf.com/>  |
| `<style>`                   | Disable all other selectors first - <https://hermitage.nl/nl>      |
| `<script>`                  | Disable all other selectors first – <https://www.truehome.com.mx/> |

There are a few other attributes where I’ve seen image URLs that could be detectable (`link[href]`, `a[href]`), but this seems quite rare so doesn’t feel worth it. `video[poster*='images/'` is also something I’ve only seen once, but it feels like it’s just a matter of sample size being too small (I mostly focused on sites’ homepages).


AliasIO:
Looks good, thank you!

thibaudcolas:
@AliasIO quick question – I see that the Wappalyzer extension got released with Wagtail detection included, which is lovely! Now clicking the "Wagtail" link in the extension goes to a missing Wagtail page on the Wappalyzer website: https://www.wappalyzer.com/technologies/cms/wagtail/.

What does it take to get that page up? I appreciate there might be additional effort on your part for this, just want to get a sense of if/when that page might be up. In the meantime I’m hesitant to announce the fact that Wappalyzer now detects Wagtail to people who build Wagtail sites. Apologies if this is documented somewhere and I missed it, and let me know if I can help.

AliasIO:
@thibaudcolas The page is there now but there's no data yet. This can take a few weeks.

https://www.wappalyzer.com/technologies/cms/wagtail/

thibaudcolas:
Thank you – I’ll wait some more before sharing and see how it evolves.

PR #3553, opened 2020-11-23
Add WebAssembly

kirisaki:
I wondered whether watching a Content-Type or a extension, but I choose the former. Because after I saw codes roughly, I noticed implement of "scripts" make browser evaluate a script. I don't know can browsers evaluate wasm byte-codes directly, but I tried it and it worked. Though it seemed strange action, so I decide to watch headers.


kirisaki:
I have remembered I found a problem, though it doesn't relate directly. But its scope is broad.

AliasIO:
Thanks!

PR #3568, opened 2020-11-29
fix: remove unused image

ayalhw:
it broke check.
the image default-safari.svg was added at https://github.com/AliasIO/wappalyzer/commit/d1f425b4b8806c668ae70aa76e9fed027762b361
and then the check was broken
Related: https://github.com/AliasIO/wappalyzer/issues/3567


AliasIO:
I've added a different fix for this, thanks.

PR #3569, opened 2020-11-29
Add pricing,plans

nurbek91:
Added prices for technologies that I previously added


AliasIO:
Thanks! Each technology should only have one of `low`, `mid` and `high`. Do you want to update those?

PR #3576, opened 2020-12-01
Add ClustrMaps/Maps category

nurbek91:
Website:
https://clustrmaps.com/

Example websiites:
http://www.equasis.org/EquasisWeb/public/HomePage
https://psyoffice.ru/
https://www.keeptalkinggreece.com/


AliasIO:
Instead of this:

 ```
       "img": {
          "attributes": {
            "src": "clustrmaps\\.com"
          }
        }
```

It's better to do this:

```
       "img[src*=clustrmaps.com]": {
          "attributes": {
            "src": ""
          }
        }
```

The first one will analyse every `img` on the page. The second one will only analyse `img` tags with `clustrmaps.com` in the `src`.

PR #3587, opened 2020-12-04
Update Tailwind CSS

ardasoyturk:


ardasoyturk:
I know "html" is deprecated but it's the only way to make this thing work. If we can do this with "dom" tag, please let me know.

PR #3592, opened 2020-12-06
Add Docusaurus to the SSG category

lex111:
Actually Docusaurus is static generator website. Of course, it is used a lot as a tool for creating documentation. But with [v2](https://v2.docusaurus.io/), which many people are already using, Docusaurus has become a full-fledged SSG.

сс @AliasIO


AliasIO:
Thanks!

PR #3605, opened 2020-12-09
Add SMSBump/Marketing automation

nurbek91:
Website:
https://smsbump.com/

Example websites:
https://www.dickpondathletics.com/ (BigCommerce )

Not covered examples (Shopify):
**https://d3dsecurity.com/, https://www.furgrip.co.uk/** 
```
"link[class*='smsbump']": {
          "attributes": {
            "class": ""
          }
        }
```

**https://lucyave.com/**
```
"div[id*='smsbump-form']": {
          "attributes": {
            "id": ""
          }
        },
```

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -14175,6 +14175,27 @@
       },
       "website": "https://www.smartstore.com"
     },
+    "SMSBump": {
+      "cats": [
+        32
+      ],
+      "description": "SMSBump is a text marketing automation app available for Shopify, BigCommerce and OpenCart.",
+      "icon": "SMSBump.svg",
+      "dom": {
+        "div[id*='smsbump-form']": {
+          "attributes": {
+            "id": ""
+          }
+        }
+      },
+      "js": {
+        "SMSBumpForm": ""
+      },
+      "scripts": "smsbump\\.com",
+      "saas": true,
+      "pricing": ["low", "freemium", "recurring"],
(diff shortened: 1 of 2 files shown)


src/technologies.json
```diff
@@ -14175,6 +14175,32 @@
...
       "website": "https://www.smartstore.com"
     },
+    "SMSBump": {
+      "cats": [
+        32
+      ],
+      "description": "SMSBump is a text marketing automation app available for Shopify, BigCommerce and OpenCart.",
+      "icon": "SMSBump.svg",
+      "dom": {
+        "div[id*='smsbump-form']": {
```
AliasIO, line 14185: Should this be a partial match or can you just use `"div.smsbump-form"`?

src/technologies.json
```diff
@@ -14175,6 +14175,32 @@
...
+          "attributes": {
+            "id": ""
+          }
+        },
+        "link[class*='smsbump']": {
```
AliasIO, line 14190: Can this be `"link.smsbump"`?

src/technologies.json
nurbek91: Yes a partial match is needed

src/technologies.json
nurbek91: I can remove DOM search. Since this is a Shopify specific problem only.
I had to go there because of the async scripts that probably do not have time to load.

PR #3607, opened 2020-12-10
Update apps.json

taggboxwappalyzer:
Updated tracking url from web.taggbox.com to widget.taggbox.com


AliasIO:
Thanks! I changed to it `(widget|web)` to capture sites using either. Please let me know if that's wrong.

PR #3617, opened 2020-12-11
Update generator regex for Docusaurus

lex111:
The meta generator value of the second (new) version of Docusaurus looks like this: `Docusaurus v2.0.0-alpha.x`.
So currently, in fact, Wapplyzer does not recognize sites created with Docusaurus v2. But without breaking the recognization of sites on Docusaurus v1 (just `Docusaurus` without specifying version).
This PR is trying to fix it.

сс @AliasIO

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -4419,7 +4419,7 @@
         "search.indexName": ""
       },
       "meta": {
-        "generator": "^Docusaurus$"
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

lex111:
@AliasIO could you please check the correctness of this regular expression? I just went to https://v2.docusaurus.io/ now and I don't see that Wappalyzer recognizing Docusaurus use. Maybe I was wrong?

PR #3672, opened 2020-12-29
fix malformed JSON

Kikobeats:
Please, next time add a way to lint your JSON file (such as https://github.com/zaach/jsonlint) before commit it!


AliasIO:
Thanks!

Issue #3687, opened 2021-01-04
Improve Shopware detection


derweili:
Shopware detection is currently not stable. 
This could be fixed by adding `/themes/Frontend/` as a rule to scripts.

AliasIO:
Do you have some examples where the detection isn't working?

derweili:
@AliasIO yes, this site: https://www.coffeecircle.com/

kyletaylored:
Looking at what the current detection is, I would say it's pretty sufficient so far. Adding `themes/Frontend` would be too generic and promote false positives. Great example, I randomly found this site (bama.com.sg) that has a `/themes/frontend` directory, but is not a Shopware site by any means. That path could be used by many different applications that utilize a `/themes` directory in the root.

For Shopware, I would say additional detection should be focused on widgets and scripts they provide to be embedded.

PR #3688, opened 2021-01-04
Add additional Shopware Script rule

derweili:
Fix for #3687


AliasIO:
Thank you. This will cause too many false positives, I can think of at least two other platforms that use this path.

PR #3710, opened 2021-01-12
Fix Oracle Commerce

### LINKED ISSUE #3667: Incorrect detection of Oracle Commerce
Example - https://www.terral03.com/

Oracle commerce is incorrectly detected as eCommerce platform.

This site seems to have link (href="https://www.bestbuy.com/site/searchpage.jsp?_dyncharset=UTF-8&id=pcat17071&iht=y&keys=keys&ks=960&list=n&sc=Global&st=wifi%20booster&type=page&usc=All%20Categories") which is messing up I think due to 'dyncharset' in the link

kyletaylored:
Remove false positive from Oracle Commerce. The dyncharset is too vague and promotes false positives. This should be detected as a URL parameter rather than a string on the page, so instead we'll remove it for now.

fixes #3667

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -11806,7 +11806,6 @@
       "headers": {
         "X-ATG-Version": "(?:ATGPlatform/([\\d.]+))?\\;version:\\1"
       },
-      "html": "<[^>]+_dyncharset",
       "icon": "Oracle.png",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

PR #3729, opened 2021-01-20
Optimizing SVG conversion build time

kyletaylored:
After running multiple builds lately for testing, the build time was taking an atrocious amount of time to run. Seeing that the bulk of the build was relying on the SVG conversion shell commands, I thought I would attempt to convert that to a Node script that was a little more efficient. Based on the timings below, we're seeing roughly a 95% decrease in build time, going from over 5 minutes to just under 14 seconds.

**Before**
```
yarn build  50.41s user 15.09s system 20% cpu 5:23.98 total
```

**After**
```
yarn build  7.99s user 2.18s system 73% cpu 13.818 total
```

- Convert bash script to Node script
- If PNG, copy over as-is, don't check
- Loop through SVGs and check if the converted file exists
- Check if SVG source file has recently changed. If it has, then convert it, else just skip it.

There's still some work to be done on efficiencies and a Node warning I'm seeing related to the promises for when SVGs are being converted, but overall it looks promising.

```
(node:2371) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(node:2371) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added to [process]. Use emitter.setMaxListeners() to increase limit
```

### DIFF
--- a/bin/convert.js
+++ b/bin/convert.js
@@ -16,64 +16,73 @@ const appPaths = () => {
   }
 }
 
+/**
+ * Copy files from source to destination.
+ * @param source
+ * @param destination
+ */
+function copyFiles(source, destination) {
+  // File destination will be created or overwritten by default.
+  fs.copyFileSync(source, destination)
+  // console.log(`${source} -> ${destination}`)
+}
+
+/**
+ * Get extension of image file.
+ * @returns {string}
+ */
+function getFileExtension(filePath) {
+  return path.extname(filePath)
+}
+
+/**
+ * Get base name of image file.
+ * @returns {string}
+ */
+function getFileName(filePath) {
+  return path.basename(filePath, getFileExtension(filePath))
+}
+
+function getConvertFileName(filePath) {
+  const name = getFileName(filePath)
+  return `${appPaths().convertPath}/${name}.png`
+}
+
+/**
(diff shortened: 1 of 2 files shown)


kyletaylored:
@AliasIO before I put anymore effort into this, just wanted to get your initial feedback on if this would be valuable to put some more work into, or not really worth it?

AliasIO:
@kyletaylored Yes, this is great. Thank you. :)

kyletaylored:
Oh no, I didn't finish it for an official pull yet! 😅

AliasIO:
Looks like it should pretty much work, right?

kyletaylored:
Technically, yes? The two main things I needed to look into further were:

1. How to detect a difference in files (this assumes something like a week)
2. Fixing the Node memory error / warnings

AliasIO:
The warning in not a concern, we can just raise the limit.

For the file diffs we can probably just generate an hash for each file and compare those. I don't think we need the date check.

AliasIO:
I optimised the script, reusing the Chromium instance helps a lot with memory and crashes. :)

PR #3730, opened 2021-01-20
Add plain/text detection

### LINKED ISSUE #2971: fail to detect CloudFlare on rawgit.loltek.net
**Platform**
Firefox 68.5.0esr on Windows 10 x64
* WebExtension (Google Chrome, Mozilla Firefox)
* Website https://rawgit.loltek.net/

**Describe the bug**
https://rawgit.loltek.net/ is using cloudflare CDN, which is supported by wappalyzer ( https://www.wappalyzer.com/technologies/cloudflare ) , but for some reason, wappalyzer is not detecting cloudflare on the site.

**To Reproduce**
1. Go to https://rawgit.loltek.net/
2. check wappalyzer

**Expected behavior**
wappalyzer says "CloudFlare CDN"

**Additional context**
what actually happens: `No technologies detected.`

kyletaylored:
While there are many MIME types that could be detected, `text/plain` should likely be added as it's a common response type from web servers, whether it's a raw file or API response.

fixes #2971

### DIFF
--- a/src/drivers/webextension/js/driver.js
+++ b/src/drivers/webextension/js/driver.js
@@ -267,13 +267,7 @@ const Driver = {
             )
           })
 
-          if (
-            headers['content-type'] &&
-            (/\/x?html/.test(headers['content-type'][0]) ||
-              /\/plain/.test(headers['content-type'][0]))
-          ) {
-            await Driver.onDetect(request.url, analyze({ headers }))
-          }
+          await Driver.onDetect(request.url, analyze({ headers }))
         }
       } catch (error) {
         Driver.error(error)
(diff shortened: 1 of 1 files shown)


AliasIO:
Let's just remove it. Headers can always be inspected.

divinity76:
@AliasIO do you mean remove the check entirely and inspect every kind of response? (rather than the just whitelisted html/plain responses) 

i don't think that's a good idea, inspecting large `Content-Type: application/octet-stream` or image/jpg or whatever could result in a significant cpu increase for little gain, i suspect the html whitelist was originally added for a good reason

AliasIO:
Only for response headers when viewing a file directly. The content script will still only run on HTML pages.

PR #3757, opened 2021-02-06
added saas key, pricing key, www. prefix to the URL, and modified the…

AbhinavS96:
Made 3 changes to ExpertRec

1. Added www. prefix to the URL
2. Added pricing and saas keys
3. Modified the script matching regex to make it more inclusive. The previous version was not matching on Shopify stores.


AliasIO:
Thanks!

PR #3758, opened 2021-02-07
Fix Akamai mPulse

kyletaylored:
**Update Akamai mPulse**
- Remove implied Akamai (not dependent upon service)
- Update detection. mPulse uses an underlying open source library called Boomerang that can be used independently from the paid service.

**Added Boomerang**
- Separated out Boomerang library from Akamai mPulse that is built on top of it.


AliasIO:
Thanks!

PR #3783, opened 2021-02-16
Change "Iubenda" to "iubenda"

giniyat202:
[The correct name is "iubenda" (with lower case "i"), not "Iubenda"](https://www.iubenda.com/en/help/158-imprintimpressum-2)


AliasIO:
Thanks!

PR #3808, opened 2021-03-03
Emergency System WordPress Plugin

DmitrysXYZ:
Emergency System WordPress Plugin

Stay up to date on all the latest developments in virology and biotechnology with ongoing updates from the W.H.O. ( World. Health. Organization ).
Helping people stay alert in the event of a biological hazard.
* You can also use our application on your own website through the wordpress plugin


DmitrysXYZ (commented):
update

AliasIO:
Thanks @DmitrysXYZ, but we generally don't include plugins as there are just too many.

PR #3819, opened 2021-03-11
Add Builder.io / CMS

teleaziz:
Website: https://builder.io

Examples:
- https://www.shopstyle.com
- https://atoms.com
- https://builder.io


AliasIO:
Thanks! I replaced the deprecated `html` with the more efficient `dom` method.

PR #3820, opened 2021-03-11
Improve Detection on Landing Pages created with Uscreen

diegoquintero-co:
The technology has been added previously, but the detection doesn't work well on Landing Pages.


AliasIO:
Thanks, I changed to a more efficient `dom` method.

PR #3822, opened 2021-03-11
Fix category name

nurbek91:
Hey mate, corrected my mistake


AliasIO:
We use British English (not US). I changed it to 'Personalisation', which is correct.

PR #3824, opened 2021-03-11
Add rules to detect HTTP authentication

Zer1t0:
Hi,

I have added some rules to detect HTTP authentication methods based on the `WWW-Authenticate` header. The methods added are:
- Basic
- Digest
- SPNEGO (Negotiate)
- NTLM
- Kerberos

Here are some examples of webs with those methods:
- https://www.raccordement-cigeo.fr/
- https://mail.gocdt.com/
- https://portal.firmade.it/
- https://devel.inexio.net/

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -2492,6 +2492,16 @@
       "scripts": "thebase\\.in/js",
       "website": "https://thebase.in"
     },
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

Issue #3830, opened 2021-03-11
React is a JavaScript library not a JavaScript framework


schmidi64:
WebExtension (Google Chrome) display React as a frameworks but it is a library.

**React - A JavaScript library for building user interfaces.** (https://reactjs.org/)

AliasIO:
Done, thanks!

PR #3832, opened 2021-03-13
udating the category of react to 59 from 12

AbhinavS96:
#3830 
Changing the category to mark react as a library.


AliasIO:
Thanks!

PR #3842, opened 2021-03-16
Add Retargeting / category

nurbek91:
Add new category Retargeting


AliasIO:
I'll do this but all locales need to be updated. It's fine to just add the English string to all files.

PR #3853, opened 2021-03-25
Updated Snowplow Analytics listing in technologies

jrpeck1989:
Updated the Snowplow Analytics listing in the technologies.json file.


AliasIO:
Thanks!

PR #3875, opened 2021-04-15
Improve Yoast SEO detection

jdevalk:
Some sites filter out the "optimized with" comment, adding the `script` tag recognition will improve detection.

Since Yoast SEO also exists for other platforms, the "implies WordPress" bit isn't true.


AliasIO:
Thanks! I updated the `html` pattern to use the new `dom` method.

PR #3990, opened 2021-06-05
Added Spaces

henrynewman:
Found their SVG logo here -> https://logos.fandom.com/wiki/Spaces


AliasIO:
Do you have some examples of websites that use it?

PR #4029, opened 2021-06-20
Update JP translate

windchime-yk:
Added, modified, and deleted Japanese translations while comparing with the latest messages.


AliasIO:
Thanks!

Issue #3957, opened 2021-05-19
jQuery detects wrong version number


qfayet:
**Platform**
Which platform does the bug apply to?
* WebExtension (Google Chrome, Mozilla Firefox)
* NPM

**Describe the bug**
On the site https://almond.consulting/, the version of jQuery is wrong.
In fact, it is corresponding to the version of jQuery UI.

**To Reproduce**
1. Go to 'https://almond.consulting/'
2. Click on 'Wappalyzer extension'
3. Scroll down to 'jQuery'
4. The version of jQuery is wrong (it should be: 3.5.1)

![image](https://user-images.githubusercontent.com/48323709/118745635-e2668580-b856-11eb-9b03-aeaf23c1c208.png)

**Expected behavior**
The version of jQuery should be: 3.5.1 (due to `.../wp-includes/js/jquery/jquery.min.js?ver=3.5.1`).

**Additional context**
The problem is that one of the regex to determine jQuery is too permissive : `jquery.*\.js(?:\?ver(?:sion)?=([\d.]+))?`
It match with: `.../wp-includes/js/jquery/ui/core.min.js?ver=1.12.1`
Maybe, it should use: `jquery[^/]*...` ?

AliasIO:
This will be fixed in the next update, thanks!

qfayet:
Great! Thanks to you!

PR #4044, opened 2021-06-22
insert back removed jquery regexes

TheConciergeDev:
The previous "scripts"-regexes (present until v6.6.0) were removed and replaced with "jquery". For me the version detection has declined due to this change. May I ask, what is the reason behind this? If I have missed something, sorry for bothering and feel free to close this proposed change. 

Thanks in advance

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -23502,7 +23502,9 @@
       "description": "jQuery is a JavaScript library which is a free, open-source software designed to simplify HTML DOM tree traversal and manipulation, as well as event handling, CSS animation, and Ajax.",
       "icon": "jQuery.svg",
       "js": {
-        "jQuery.fn.jquery": "([\\d.]+)\\;version:\\1"
+        "jQuery.fn.jquery": "([\\d.]+)\\;version:\\1",
+        "$().jquery": "([\\d.]+)\\;version:\\1",
+        "$.fn.jquery": "([\\d.]+)\\;version:\\1"
       },
       "scripts": [
         "jquery[.-]([\\d.]*\\d)[^/]*\\.js\\;version:\\1",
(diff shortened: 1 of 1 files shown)


AliasIO:
It's because of #3957, the version was often wrong. The jQuery version can usually be detected through `jQuery.fn.version`, which is always accurate.

TheConciergeDev:
Hi,

thanks for your reply. I would like to challenge this change a little bit (in a positive way). I tested this now on a few different pages.

With `jQuery.fn.version` only, I am not able to detect the the jQuery version of amazon.com, stackoverflow.com, nor the `qfayet` website (referenced in #3957)

I even tried different js-code snippets (found [here](https://stackoverflow.com/questions/55722181/jquery-version-check/55722259#55722259))
* `$().jquery`
* `$.fn.jquery`

Checks:
* amazon.com - the html regexes helped me, the snippets did not (seems to be kind of a "custom" implementation)
* stackoverflow.com - different js snippets helped, determining the version + regexes too.
* github.com - `jQuery.fn.version` returns as well: `undefined`. The other snippets helped here too + regex got it right as well.

I would now extend the `js`-detection + leave the html regexes as a backup.

What are your thoughts on this?

Greetings & thanks in advance

AliasIO:
We can add `$.fn.jquery` but `$().jquery` won't work as we only analyse static variables.

For the `scripts` patterns we can use `(\\d+\\.\\d+\\.\\d+)` to ensure the format is correct. I don't trust the `?version=xxx` parameter but these would work: 

```
      "js": {
        "jQuery.fn.jquery": "([\\d.]+)\\;version:\\1"
        "$.fn.jquery": "([\\d.]+)\\;version:\\1"
      },
      "scripts": "jquery",
      "scripts": [
        "/jquery(-(\\d+\\.\\d+\\.\\d+))[/.-];version:\\1",
        "/(\\d+\\.\\d+\\.\\d+)/jquery[/.-];version:\\1"
      ],
 ```

TheConciergeDev:
Thanks for your reply,

I agree `?version=XXX` is too generic and can be ignored.  I hope the change is okay for you. 

In your comment you put `"scripts"` twice in there - I guessed a copy issue, therefore I only added the list.

AliasIO:
Thanks!

PR #4049, opened 2021-06-23
Ghost is a CMS

jdevalk:
Ghost calls itself a publishing platform, so calling it just a blog is not really correct.


AliasIO:
Thanks!

PR #4053, opened 2021-06-24
Upd ip-api ico to 32x32,Upd Fingerprint version

nurbek91:
https://dev.fingerprintjs.com/docs/pro-vs-open-source

examples:
https://people.cs.kuleuven.be/~victor.lepochat/mwe.html
https://uk.wallapop.com/
https://resolume.com/
https://www.as-web.jp/
https://www.comuniazo.com/


VictorLeP:
I see you hardcoded `dev` in the regex. However, there's for example also `beta` (see `3.0.0-beta.3` in [this list](https://www.npmjs.com/package/@fingerprintjs/fingerprintjs)).
Is there no way to create the regex such that it just accepts any character other than `/` in the version number? (so without the need to hardcode dashes/dots/etc.)
The strange thing is that the regex I wrote seems to work just fine at https://regex101.com/ (which is why I can't really grasp why the regex fails here / develop a working version), but the variant regex processor used here seems to be thrown off by the dash.

nurbek91:
@VictorLeP so that it takes every possible set of alphabetic characters at the end.
I think there is already a limitation of the Wappalyzer itself. I don't know much about these intricacies. Yes, it came to omit the symbol -

VictorLeP:
My original regex is actually completely fine (i.e., dashes are no issue). 
The detection fails because version numbers are limited to 10 characters:
https://github.com/AliasIO/wappalyzer/blob/8b1f03e8da5680d0c8842442356fc4a0fcbfbbc1/src/wappalyzer.js#L52
So a fake example of version `3.1.0-d.0` works fine because it passes the length check; conversely, an arbitrary long number (`3125326506506`) fails even though it passes the regex.

So, the fix would be to increase the allowed length for a version number (if that is OK of course), and then restore the version detection regex `(\\d.*?)` in order to cover all future (edge) cases. 
@nurbek91: I think your cleanup of the other parts (`(?:\\-pro|2)?(?:@|/)`) are indeed an improvement, so a hybrid of the two would be awesome.

AliasIO:
The limit is mainly there to avoid strings like timestamps and hashes. I'll raise the limit to 15 and add an additional check for long numeric strings:

```
_version.length <= 15 && (parseInt(_version, 10) || 0) < 10000
```

VictorLeP:
Thanks for increasing the length limit on version numbers.

This PR wasn't meant to be merged yet, as I had [commented](https://github.com/AliasIO/wappalyzer/pull/4053#issuecomment-868276095) that the regex was still incorrect. I've created a new PR (#4088) which I think improves and corrects the regex.
@nurbek91 could you please verify the new regex? I had to add a `/` at the end to make the version detection work on some sites (e.g., https://www.drtusz.pl/), it used to work with my version where I had separate regexes, but your merged version broke it.

nurbek91:
Hey @VictorLeP yes confirm updates on all examples here  https://regex101.com/r/7rD1mn/1 Thanks!!

PR #4116, opened 2021-07-03
fix: imbalanced 'localised' dom pattern

kingthorin:
Pproperly match square brackets.

Introduced in: https://github.com/AliasIO/wappalyzer/pull/4001/

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>


AliasIO:
Thanks!

PR #4120, opened 2021-07-04
Updated Korean translation

ddarkr:
Hello, I updated korean translation.
Some direct translations because there are many parts can't be checked.


AliasIO:
Thanks!

PR #4151, opened 2021-07-07
Improve Apollo by detecting script with state

yuriy636:
As seen on this site: https://www.eneba.com/

And it's documented in the docs: https://www.apollographql.com/docs/react/performance/server-side-rendering/#executing-queries-with-getdatafromtree


AliasIO:
Thanks! I updated it to use `dom` instead of `html`.

PR #4222, opened 2021-07-20
Update Zendesk categories

kyletaylored:
Zendesk is not actuall a CMS, but a knowledge base and support ticket platform, which causes data issues that conflict with actual CMSs. The detection is geared towards JS, which infers chat and widgets, which make sense. Reassigning this as a documentation and and issue tracker, which seem more appropriate.


AliasIO:
Thanks!

PR #4233, opened 2021-07-20
Fix Duda detection

kyletaylored:
- Removed arbitrary HTML (causing false positives)
- Removed vague JS object
- Added Duda specific library detection
- Added Duda description
- Changed category to landing page builder (Duda is not a CMS)
- Change GoDaddy Website Builder category to page builder as well (it's not a CMS)

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -13580,13 +13580,11 @@
     },
     "MyWebsite": {
       "cats": [
-        1
+        51
       ],
-      "description": "MyWebsite is website builder designed for easy editing and fast results.",
-      "icon": "Ionos-by-1and1-logo.svg",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

liady:
Hi @kyletaylored, @AliasIO , Liad from Duda here - 
I see that the new changes made in this PR are not actually correct to identify Duda platforms.
We use the `id` as a stable way of identification, and we do not rely on the JS parameters (they can be easily changed).
I actually merged a PR few days before this PR (#4189 ) that sets Duda's recognition in the correct way.
In addition - Duda is a CMS, not a landing page builder.

Can you please revert this PR? It is wrong.

PR #4244, opened 2021-07-26
Change Builder to Builder.io

teleaziz:
We've been using Builder and Builder.io interchangeably in our online communications, we'd like to move to just using `Builder.io` , thank you


AliasIO:
Thanks!

Issue #4303, opened 2021-08-06
Add Category For WordPress Hosting


noodles:
**Is your feature request related to a problem? Please describe.**
There are a number of WordPress plugins already present in technologies.json
I think it would be useful to have a separate category for WordPress hosting.

**Describe the solution you'd like**

1. Create a new category for WordPress hosting
2. Add this new category to the exisiting WordPress hosting in technologies.json 

**Describe alternatives you've considered**
Don't create a category for WordPress Hosting Platforms :)

**Additional context**

List of WP plugins currently in technologies.json:

- [x] WP Engine
- [x] Flywheel
- [x] Kinsta
- [x] Pagely
- [x] Seravo

PR #4304, opened 2021-08-06
Add Category For WordPress Hosting & Update Exisiting Entries

noodles:
As detailed in this FR:

Add Category For WordPress Hosting #4303


noodles:
Does it make sense to switch these WP hosting entries to use `requires` instead of `implies`

AliasIO:
That depends. Should this perhaps be a generic 'hosting' category?

noodles:
The ones listed in #4303 are all WordPress only hosting so it's big enough to be a standalone category BUT thinking about how it would appear in the plugin, just "hosting" might make more sense.

noodles:
Would just `hosting` be OK for the category name?

    "88": {
      "name": "Hosting",
      "priority": 9
    }

AliasIO:
Yes, let's go with that. That way we can add other web hosts as well.

PR #4335, opened 2021-08-09
Fixed the identification for Hubspot consent functionality

max-ostapenko:
See discussion here: https://github.com/HTTPArchive/almanac.httparchive.org/issues/2292

Before: JS variable and request mentioned were always available for HubSpot Analytics tech.
Now: querying for the actual node of the banner.

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -9951,13 +9951,10 @@
       "cats": [
         67
       ],
-      "description": "HubSpot Cookie is a cookie compliance tool from HubSpot.",
+      "description": "HubSpot Cookie Policy banner is a cookie compliance functionality from HubSpot.",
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks. Should we leave the `js` and `scripts` methods for older implementations or are they not relevant at all?

max-ostapenko:
This fix was exactly about removing older implementation and a distinct functionality with the correct definition.
So no, js and script are currently irrelevant.

max-ostapenko:
Maybe at the time when this definition was added it was correct.
But currently, it looks like the identification methods leading to duplicates: HubSpot Analytics and HubSpot Cookie.

AliasIO:
Makes sense, thanks!

PR #4366, opened 2021-08-11
Fixed 'Salesforce commerce cloud'  / 'Orckestra' / 'Borderfree' / Added 'Reviews' and 'Translation' categories

rockeynebhwani:
Updated for following -

- Fixed for 'Salesforce commerce cloud' false positives (https://github.com/AliasIO/wappalyzer/issues/4090)
- Improved detected for 'Orckestra' (https://github.com/AliasIO/wappalyzer/issues/4362) and 'Borderfree' (https://github.com/AliasIO/wappalyzer/issues/4356)

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -18496,8 +18496,7 @@
       "icon": "Salesforce.svg",
       "implies": "Salesforce",
       "js": {
-        "dwAnalytics": "",
-        "dw": ""
+        "dwAnalytics": ""
       },
       "pricing": [
         "poa"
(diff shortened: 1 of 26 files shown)


src/technologies.json
```diff
@@ -18473,17 +18485,19 @@
...
       },
-      "html": [
-        "<[^>]+demandware\\.edgesuite"
-      ],
       "icon": "Salesforce.svg",
       "implies": "Salesforce",
       "js": {
-        "dwAnalytics": ""
+        "dwAnalytics": "",
+        "dw": ""
```
AliasIO, line 18500: Very short variables like these are often found in minified scripts, causing false positives. I think we can use `dw.ac` in this case.

rockeynebhwani: Removing 'dw' variable. 'dwAnalytics' should be enough to identify anyways based on my knowledge of this platform.

AliasIO:
Thanks!

PR #4375, opened 2021-08-12
Improve FancyBox version detection

jkmartindale:
`jQuery.fancybox.version` for when `$.fancybox` is undefined  
`Fancybox.version` for v4 detection


AliasIO:
Thanks!

PR #4471, opened 2021-08-20
Add Automattic / Newspack

adamsilverstein:
Vendor URL: https://newspack.pub/  
Description: Automattic's Newspack is an all-in-one platform designed for small and mwedium-sized news organizations that simplifies publishing and drives audience and revenue right out of the box.
Icon: newspack.png
Pricing: mid, recurring

Example URLs:
1. https://www.elsoberano.org/	
2. https://www.hechingerreport.org/	
3. https://www.oaklandside.org/	
4. https://www.calmatters.org/	
5. https://www.blackvoicenews.com/

Check for:
- header: `host-header=Newspack`
- a href: `newspack.pub` + `class='imprint'`


src/technologies.json
```diff
@@ -14376,6 +14376,28 @@
...
+        "host-header": "Newspack"
+      },
+      "icon": "newspack.png",
+      "dom": {
+        "link[href*='newspack.pub']": {
+          "attributes": {
+            "class": "imprint"
+          }
+        }
+      },
```
AliasIO, line 14394: Should this be `"dom": "a.imprint[href*='newspack.pub']"`?

adamsilverstein: Right! yes, I will fix this PR up based on your feedback on the Newspack Plugin PR.

AliasIO:
Thanks!

src/technologies.json
adamsilverstein: @AliasIO I wound up removing the `dom` check in favor of the `headers` check. In my testing when I had the dom check in place, sites that had the plugin but were not hosted by Automattic (and thus did not server the "host-header": "Newspack") were still identified as Newspack by Automattic sites. With only the header check, everything works as expected.

This PR works as expected for me now, however I'm still a little confused :)

Since I'm new around here I'll ask this basic question: is that expected? that is do the "fingerprints" work by matching **any** of the specified items, or should it require matching **all** conditions? 

I also tried making Newspack Hosting **require** Newspack (instead of **implies**) thinking this would keep the dom check in the plugin fingerprint but still make it required for this one. When I tried this however, the fingerprint failed to identify my test site as. Newspack by Automattic site.

For reference, I'm testing with the following sites:
_https://www.elsoberano.org/_ - Newspack by Automattic hosting AND Newspack plugin. 
_https://cowgirlmagazine.com/_ - Only the Newspack plugin (not hosted by automattic, no "host-header": "Newspack")

adamsilverstein:
@AliasIO Thanks for merging!

I wound up removing the dom check in favor of the headers check. In my testing when I had the dom check in place, sites that had the plugin but were not hosted by Automattic (and thus did not server the "host-header": "Newspack") were still identified as Newspack by Automattic sites. With only the header check, everything works as expected.

This PR works as expected for me now, however I'm still a little confused :)

Since I'm new around here I'll ask this basic question: is that expected? that is do the "fingerprints" work by matching any of the specified items, or should it require matching all conditions?

I also tried making Newspack Hosting require Newspack (instead of implies) thinking this would keep the dom check in the plugin fingerprint but still make it required for this one. When I tried this however, the fingerprint failed to identify my test site as. Newspack by Automattic site.

For reference, I'm testing with the following sites:
https://www.elsoberano.org/ - Newspack by Automattic hosting AND Newspack plugin.
https://cowgirlmagazine.com/ - Only the Newspack plugin (not hosted by automattic, no "host-header": "Newspack")

AliasIO:
We look for _any_ matching pattern. You can append `\\;confidence:50` to two patterns if you want both to match.

If you're testing in the browser extension, be sure to clear the detections cache in the extension settings between changes or disable the cache completely. Otherwise detections will be remembered even when a pattern has been removed.

adamsilverstein:
Thanks for clarifying and explaining the `confidence` variable. 

I'm testing by running `node src/drivers/npm/cli.js testdomain.com` and learned to run `npm run link` after making changes. Seemed to pick up changes every time that way.

Thanks again for such an excellent resource for the web, this tool is so important I'm really  grateful you are maintaining it. Thank You!

ps. I'll bet you never expected technologies.json to grow to 30k lines! some day it might make sense to break that up, my editor quietly complained when I opened it :)

AliasIO:
Thanks Adam!

Are you on Windows perhaps? `yarn run link` is supposed to create a symlink that keeps the files the same.

And yes, I'll look into splitting the file up. It's now so large that I can't PRs directly on GitHub anymore.

AliasIO:
I split the `technologies.json` file up into `technologies/[a-z].json` in 2401457a994f40f23140b62e406590bdaa0476a2. See [/src/technologies](https://github.com/AliasIO/wappalyzer/tree/master/src/technologies).

cc @nurbek91 @pmeenan @rockeynebhwani @noodles

PR #4483, opened 2021-08-22
Fix Duda detection

liady:
Duda uses the id as a stable way of identification, and do not rely on the JS parameters (they can be easily changed).
In addition - Duda is indeed a CMS, not a landing page builder.


AliasIO:
Thanks!

PR #4421, opened 2021-08-16
Add core-js

jkmartindale:
15.1k stars on GitHub: https://github.com/zloirock/core-js  
Example website: https://www.bbc.com/

PR #4505, opened 2021-08-23
Fix `core-js` detection

zloirock:
In #4421 was added `core-js` detection, but it misses the most of cases.

The global `core` variable was available to `core-js@2` in *some* cases.

The global `__core-js_shared__` property is available in modern versions in *most* cases.

The `.version` and `.versions` properties are available only in modern versions.

`window['__core-js_shared__'].versions` contains information about all loaded `core-js` versions and modes, but I have no ideas how to handle it with your regex syntax.

image

My detection script https://github.com/zloirock/core-js/blob/5ddafd1/scripts/usage.mjs

Top 1000 results https://gist.github.com/zloirock/7331cec2a1ba74feae09e64584ec5d0e

Top 10000 results https://gist.github.com/zloirock/1fae0b9798b50d3551432ebf100216b9

### DIFF
--- a/src/technologies.json
+++ b/src/technologies.json
@@ -5618,7 +5618,8 @@
         "__core-js_shared__": "",
         "__core-js_shared__.versions.0.version": "^(.+)$\\;version:\\1",
         "core": "",
-        "core.version": "^(.+)$\\;version:\\1"
+        "core.version": "^(.+)$\\;version:\\1",
+        "_babelPolyfill": ""
       },
       "oss": true,
       "website": "https://github.com/zloirock/core-js"
@@ -16606,8 +16607,7 @@
       ],
       "icon": "polyfill.svg",
       "scripts": [
-        "^https?://cdn\\.polyfill\\.io/",
-        "/polyfill\\.min\\.js"
+        "^https?://cdn\\.polyfill\\.io/"
       ],
       "website": "https://polyfill.io"
     },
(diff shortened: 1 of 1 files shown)


src/technologies.json
```diff
@@ -5615,7 +5615,11 @@
       ],
       "description": "core-js is a modular standard library for JavaScript, with polyfills for cutting-edge ECMAScript features.",
       "js": {
-        "core.version": "^(.+)$\\;version:\\1"
+        "__core-js_shared__": "",
+        "__core-js_shared__.versions.0.version": "^(.+)$\\;version:\\1",
+        "core": "",
+        "core.version": "^(.+)$\\;version:\\1",
+        "_babelPolyfill": ""
```
zloirock, line 5622: `babel-polyfill` was just a wrapper for `core-js`, https://github.com/babel/babel-archive/blob/15c88ef85657cd1d320e515189caebd5e26c27c3/packages/babel-polyfill/src/noConflict.js

src/technologies.json
```diff
@@ -16603,8 +16607,7 @@
       ],
       "icon": "polyfill.svg",
       "scripts": [
-        "^https?://cdn\\.polyfill\\.io/",
-        "/polyfill\\.min\\.js"
```
zloirock, line 16607: `polyfill.min.js` is not something unique for `polyfill.io` - files with this name was in many polyfills project, including old `core-js` versions.

AliasIO:
Thank you for this!

> `window['__core-js_shared__'].versions` contains information about all loaded `core-js` versions and modes, but I have no ideas how to handle it with your regex syntax.

`__core-js_shared__.versions[0].version` should work, I updated this line.

PR #4517, opened 2021-08-27
Update JP translate

windchime-yk:
The English in the Japanese translation file has been fixed to Japanese.


AliasIO:
Thank you!

PR #4518, opened 2021-08-27
Add Vue-Storefront

bloodf:
Project URL: https://www.vuestorefront.io/

Vue Storefront 2 Example URL:

https://www.vans.ca/en-ca
https://puredailycare.com/

Vue Storefront 1 Example URL:

https://zadig-et-voltaire.com/


src/technologies.json
```diff
@@ -23647,6 +23647,33 @@
       ],
       "website": "https://vuejs.org"
     },
+    "Vue-Storefront 1": {
+      "cats": [
+        6,
+        12
+      ],
+      "description": "Vue Storefront is the world's fastest-growing open-source eCommerce frontend.",
```
AliasIO, line 23655: Please change this to something more neutral, e.g. "Vue Storefront is a frontend platform for headless commerce."

AliasIO:
Thanks for the PR! Could we combine the two into a single technology and capture the version number as `1` or `2`?

E.g. `"html": "<[^>]+\\sdata-v(?:ue)?-\\;version=2"`.

bloodf:
> Thanks for the PR! Could we combine the two into a single technology and capture the version number as `1` or `2`?
> 
> E.g. `"html": "<[^>]+\\sdata-v(?:ue)?-\\;version=2"`.

The main problem is that Version 1, implies only Vue.js, and Version 2, implies Nuxt. That's why I didn't made just a single version :(

src/technologies.json
bloodf: I've updated

AliasIO:
> The main problem is that Version 1, implies only Vue.js, and Version 2, implies Nuxt. That's why I didn't made just a single version :(

What if you leave off the Nuxt.js implies? I think in most cases we identify Nuxt.js anyway.

bloodf:
@AliasIO fixed the versioning detection

AliasIO:
Perfect, thanks!

PR #4540, opened 2021-09-01
Added 'Suiteshare', 'Zoominfo chat', 'Attraqt', 'Gumstack' & 'Cloudflare bot management' / Fixed 'Wheelio' / Recategorised 'Zonos'

rockeynebhwani:
https://github.com/AliasIO/wappalyzer/issues/4302
https://github.com/AliasIO/wappalyzer/issues/4290
https://github.com/AliasIO/wappalyzer/issues/4550
https://github.com/AliasIO/wappalyzer/issues/4549
https://github.com/AliasIO/wappalyzer/issues/4542
https://github.com/AliasIO/wappalyzer/issues/4541

PR #4552, opened 2021-09-02
Update technologies.json

max-ostapenko:
There was a web API's event mapped to Wheelio app.
So this app is identified by mistake on every site.

https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent


max-ostapenko:
Duplicate of #4540

PR #4587, opened 2021-09-07
Updated Vue Storefront detection

bloodf:
update the Vue Storefront detection, so it can be widely detected.

### DIFF
--- a/src/technologies/v.json
+++ b/src/technologies/v.json
@@ -479,7 +479,7 @@
     "saas": true,
     "website": "https://www.voracio.co.uk"
   },
-  "Vue-Storefront": {
+  "Vue Storefront": {
     "cats": [
       6,
       12
@@ -488,10 +488,10 @@
     "icon": "vue-storefront.svg",
     "implies": "Vue.js",
     "meta": {
-      "generator": "^Vue Storefront\\ [0-9.]+?$;version=2"
+      "generator": "^Vue Storefront\\ [0-9.]+?$"
     },
     "scripts": [
-      "(?:\\/([\\d.]+))?\\/vsf-layout;version=1"
+      "vsf-layout(?:-\\w*)?\\.?(?:[0-9a-fA-F]{7,40}|[\\d]+(?:.[\\d]+(?:.[\\d]+)?)?)(?:\\.min)?\\.js"
     ],
     "website": "https://www.vuestorefront.io/"
   },


src/technologies/v.json
```diff
@@ -488,10 +488,10 @@
     "icon": "vue-storefront.svg",
     "implies": "Vue.js",
     "meta": {
-      "generator": "^Vue Storefront\\ [0-9.]+?$;version=2"
+      "generator": "^Vue Storefront\\ [0-9.]+?$"
```
AliasIO, line 491: Should we include the version number in this case? I noticed the old code was wrong, it should be:

`"generator": "^Vue Storefront\\ ([0-9.]+)?$\\;version=\\1"`

src/technologies/v.json
```diff
@@ -488,10 +488,10 @@
...
     },
     "scripts": [
-      "(?:\\/([\\d.]+))?\\/vsf-layout;version=1"
+      "vsf-layout(?:-\\w*)?\\.?(?:[0-9a-fA-F]{7,40}|[\\d]+(?:.[\\d]+(?:.[\\d]+)?)?)(?:\\.min)?\\.js"
```
AliasIO, line 494: Could this simply be `vsf-layout\\;version=1`? Is this used in versions other than 1?

bloodf: I tried, but the extension was not recognizing VSF. If I add the version happens the same.

src/technologies/v.json
bloodf: Adding the version, the extension does not recognize VSF

AliasIO: Sorry, the correct syntax is "^Vue Storefront ([0-9.]+)?$\\;version:\\1".

AliasIO:
I updated this in to work with version detection, see:

https://github.com/AliasIO/wappalyzer/commit/2101abddda98ba5d1cc09c806c4f24f6be262c42#diff-34996dcc0286345b8ace13f89663bf0bfdaf8ee97dc72a18dcc3124bd57a062e

bloodf:
@AliasIO the name need to be updated to `Vue Storefront` removing the dash.

Also the current version 6.8.5 is not detecting the integration :(

AliasIO:
I'll update that. It's not working in v6.8.5 because the version syntax was wrong. A new version is currently pending approval, Vue Storefront will be detect correctly in the version after that.

PR #4589, opened 2021-09-08
Fix regex brackets for MaxMind

VictorLeP:
The brackets used in the regex for MaxMind were wrong: they would match the _characters_ from the words `device`/`js`, not the full word.


AliasIO:
Thanks!

PR #4627, opened 2021-09-14
Adding Acquia products (and Drupal multisite)

kyletaylored:
Acquia has a number of extended products outside of hosting, including personalization, CDP, and multisite support.

Fixed two additional things:
- Duda false positive HTML detection
- ThreatMetrix icon was broken


AliasIO:
Thanks!

PR #4644, opened 2021-09-16
Implement actual Preact detection

developit:
The detection added in #4056 was looking for [preact-helmet](https://www.npmjs.com/package/preact-helmet), a very specific library that is not a proxy for Preact usage.


AliasIO:
Thank you!

PR #4646, opened 2021-09-16
Fix Drupal Multisite (false positive)

kyletaylored:
Fixing a false positive in the Drupal multisite detection (all D7 sites use sites/all path, marked for exclusion).


AliasIO:
Thanks!

PR #4689, opened 2021-09-23
Update managed WordPress providers

kyletaylored:
- Add WordPress VIP, WordPress.com, Pressable
- Remove Automattic

Automattic's primary business was WordPress.com, and WordPress VIP for enterprise-grade hosting. Since then, their business has expanded into multiple revenue streams and categories, everything from hosting to plugins, they also now own Tumblr. With that in mind, this will add separate detection for WordPress.com and WordPress VIP, and remove Automattic itself as it's not really a platform anymore.

This also adds another managed WordPress provider, Pressable. And ran `yarn build` to clean up the files for future merging.


AliasIO:
Thank you!

PR #4691, opened 2021-09-24
[OpenBSD httpd] Exclude Apache from detection

### LINKED ISSUE #3113: OpenBSD httpd mistakenly identified as Apache
**Platform**
WebExtension (Google Chrome, Mozilla Firefox)

**Describe the bug**
When visiting websites hosted using OpenBSD httpd, the extension shows both Apache and OpenBSD httpd as the web servers. 

image

**To Reproduce**
1. Go to https://www.openbsd.org/
2. Open the Wappalyzer extension

**Expected behavior**
Apache should not appear in the list. OpenBSD httpd is the only web server present

**Additional context**
The OpenBSD httpd web server string is `OpenBSD httpd`, which might look similar enough to the Apache rule to match it.

dantecatalfamo:
Closes #3113

### DIFF
--- a/src/technologies/o.json
+++ b/src/technologies/o.json
@@ -466,6 +466,8 @@
     "headers": {
       "Server": "^OpenBSD httpd"
     },
+    "excludes": "Apache",
+    "oss": true,
     "website": "https://man.openbsd.org/httpd.8"
   },
   "OpenCart": {


AliasIO:
Thank you!

PR #4708, opened 2021-09-29
improve jsf detection

Yofial:
Some example of detections at [publicwww.com](https://publicwww.com/websites/%3Cinput+type%3D%22hidden%22+name%3D%22javax.faces.ViewState%22/)


Yofial:
here is an example of a new detection from the previous list: [https://www.jobbank.gc.ca/home](https://www.jobbank.gc.ca/home)

AliasIO:
Thanks!

PR #4709, opened 2021-09-30
Add GoAnywhere detection

Yofial:
GoAnywhere is an MFT software.
[Gartner Peer Insights](https://www.gartner.com/reviews/market/application-development-integration-and-management-others/vendor/helpsystems/product/goanywhere-managed-file-transfer) has currently 52 reviews of the product.

The program requires login. You can find some login pages by searching google for the text "Powered by GoAnywhere" which appears on the login page.


Yofial:
The validation failure is on the s.json file which was not changed in this commit. It appears under "Unchanged files with check annotations".
This seems to be an problem with the way the Github Action is configured rather than my commit

AliasIO:
Thanks!

PR #4712, opened 2021-10-01
fix: removing implies from stitches because it can work with any js FW

mzaien:
Sorry for my miss understanding but now i know what implies meant and this will fix the detection from being wrong like this 👇
CleanShot 2021-10-01 at 15 38 02@2x


AliasIO:
Thanks!

mzaien:
Hello @AliasIO , sorry for bothering , i think there is some cache that  did not changed since this pr was merged 

still the plugin shows all implies technologies 
CleanShot 2021-12-11 at 15 08 29@2x
example :
https://ped.ro/
https://uiw.tf/
https://rauno.me/

AliasIO:
Can you try clearing the detection cache in the extension settings?

mzaien:
> Can you try clearing the detection cache in the extension settings?

I did that, and tried viewing the sites from different browsers

AliasIO:
I just noticed that your PR adds a duplicate block for Stiches instead of update what was there already. Would mind creating a new PR?

mzaien:
Sorry for making this horrible error,
i made the new pr fixing all of that no duplicate , and no implies

PR #4729, opened 2021-10-04
Add Shopify to CMS category

alonkochba:
Shopify is defined as a CMS in [w3tech](https://w3techs.com/technologies/history_overview/content_management) and also defines itself as a CMS for example - https://www.shopify.com/tour/ecommerce-cms

We will probably include it in the CMS chapter for this year's [web almanac](https://almanac.httparchive.org/en/2020/cms)

### DIFF
--- a/src/technologies/s.json
+++ b/src/technologies/s.json
@@ -1327,6 +1327,7 @@
   },
   "Shopify": {
     "cats": [
+      1,
       6
     ],
     "cookies": {


AliasIO:
Thank you. Most ecommerce platforms have CMS features but you wouldn't use them as an alternative to Drupal or Joomla. I think it makes sense to just keep them in the ecommerce category.

PR #4996, opened 2021-10-29
added a description and changed the URL for Addi

AbhinavS96:
Forgot to add a description when adding the detection. Also noticed that their website redirects to a different URL, so updated that as well.


AliasIO:
Thanks!

PR #4997, opened 2021-10-30
feat: update vercel server header

dunklesToast:
Vercel also uses the `server: Vercel` header.

Check it via: `curl -sI https://vercel.com/ | grep server:`


AliasIO:
Thanks!

PR #5022, opened 2021-11-01
added js object based detection for expertrec

AbhinavS96:
Adding js object based detection. I believe this is more efficient than scriptsrc. Keeping scriptsrc as a fallback.


AliasIO:
👍

PR #5039, opened 2021-11-01
Fix Vercel category

kyletaylored:
Vercel was previously under the "web server" category, but they are actually a Platform that runs Next.js servers.


AliasIO:
Thanks!

PR #5061, opened 2021-11-04
Update cart functionality

myfluxi:
Cart functionality is too important to be missed.

"shopping-bag/basket/cart" frequently appears in various tags, so use it.

Example: https://www.bike24.de

### DIFF
--- a/src/technologies/c.json
+++ b/src/technologies/c.json
@@ -342,7 +342,19 @@
       6
     ],
     "description": "Websites that have a shopping cart or checkout page, either using a known ecommerce platform or a custom solution.",
-    "dom": "a[href*='/cart'], a[href*='/order'], a[href*='/basket'], a[href*='/trolley'], a[href*='/bag'], a[href*='/shoppingbag'], a[href*='/checkout'], [aria-controls='cart']",
+    "dom": [
+      "a[href*='/cart']",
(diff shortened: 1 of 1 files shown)


rockeynebhwani:
@myfluxi  - My concern is that this may result in false positives.

myfluxi:
I understand the possible issue with false positives but I would give it a try. Having a cart/checkout is more or less the definition of eCommerce and the info is too valueable to be missed.

We could limit this to class attributes but we would then miss a lot of hits.

AliasIO:
@myfluxi We already capture Cart Functionality when GA Enhanced eCommerce is detected, we just don't display this in the browser extension. You can verify it [here](https://www.wappalyzer.com/lookup/bike24.de) or using the NPM driver.

Screen Shot 2021-11-06 at 6 55 09 am

myfluxi:
I guess this result is cached, Bike24 was relaunched a few days ago. There is no jQuery any more etc etc.

I could limit lookup to DOM, *[class*='shopping-cart'], *[class*='shopping-bag'], *[class*='shopping-cart'] - no false-positives.

AliasIO:
That would be better, the `html` pattern isn't as efficient.

AliasIO:
Thanks!

Issue #4937, opened 2021-10-22
Track 'Solvyy'


rockeynebhwani:
Vendor Site - https://solvvy.com/
Example Site - https://madeincookware.com/collections/shop-all
Category - Live Chat
Detection - Check for script from domain '//cdn.solvvy.com/' AND JS object 'Solvvy'

PR #5087, opened 2021-11-06
added solvvy

AbhinavS96:
Added detection based on #4937
The sample site provided (https://madeincookware.com/collections/shop-all) doesn't seem to be using this anymore.


AliasIO:
Thanks!

Issue #4795, opened 2021-10-10
Track 'Qualaroo'


rockeynebhwani:
Vendor Site - https://qualaroo.com/
Example Site - https://www.theknot.com/
Detection - Check for script from domain '//cl.qualaroo.com/'

PR #5099, opened 2021-11-07
added qualaroo

AbhinavS96:
#4795
The example site mentioned is not using this anymore. Found another: https://www.alllaw.com/


AliasIO:
👍

Issue #4788, opened 2021-10-10
Track 'Yellow.ai'


rockeynebhwani:
Vendor Site - https://yellow.ai/
Example Site - https://www.duroflexworld.com/products/back-magic
Category - Live Chat
Detection - Check for scripts from domain '//cdn.yellowmessenger.com' AND JS object 'ymConfig'

PR #5102, opened 2021-11-07
added yello ai

AbhinavS96:
#4788
Example Site - https://www.duroflexworld.com/products/back-magic


AliasIO:
Thanks! I'll replace the icon with an SVG.

Issue #5142, opened 2021-11-09
Track 'WorldShopping'


rockeynebhwani:
Vendor Site - https://www.worldshopping.global/
Example Site - https://www.ec-store.net/
Category - Ecommerce
Detection - Check for script from domain '//checkout-js.worldshopping.jp' and extract version number from URL

PR #5144, opened 2021-11-09
added worldshopping

AbhinavS96:
#5142
Added WorldShopping
Example Site - https://www.ec-store.net/

It is detecting the version as 3_5_0. How should I modify the regex to get it to show as 3.5.0?


nurbek91:
Hey there , something like this:
`\\.worldshopping\\.jp/checkout-(\\d+)_(\\d+)_(\\d+)\\.js\\;version:\\1.\\2.\\3`
We need to improve it somehow so that other options are taken:  3_5 and  3_5_5_1
I could not test locally because the script is not loaded due to the fact that it has async in its attributes. And now I'm not sure if it's possible to take a version at all.
Maybe I have problems with the Internet. Try it for yourself.

AbhinavS96:
Thanks, @nurbek91 
I have updated the regex. The detection does not work from the CLI but the web extension is picking up the version correctly. 

We will need a few more sample cases to test out options like 3_5 and 3_5_5_1 (assuming versions like that are possible). I believe that the current regex will at least detect the technology if the versioning format is different.

AliasIO:
The script URL looks like this:

```html
<script type="text/javascript" charset="UTF-8" async="true" src="https://checkout-api.worldshopping.jp/v1/script?token=ec-store_net"></script>
```

Wouldn't the version number just be `v1`? Where did you see `3_5_0`?

AbhinavS96:
The network call is showing `https://checkout-js.worldshopping.jp/checkout-3_5_0.js` 
That's where I took the version from.

nurbek91:
Hey @AbhinavS96 
> Thanks, @nurbek91 I have updated the regex. The detection does not work from the CLI but the web extension is picking up the version correctly.
> 
> We will need a few more sample cases to test out options like 3_5 and 3_5_5_1 (assuming versions like that are possible). I believe that the current regex will at least detect the technology if the versioning format is different.
A few tips.
1) It is advisable to use the (?:...)
A non-capturing group allows you to apply quantifiers to part of your regex but does not capture/assign an ID
image 
Then we are almost guaranteed to determine the technology even if the regular expression has some problems further with the version
2) you can also test without examples virtually on services such as https://regex101.com/r/waPdEx/1/

AliasIO:
`scriptSrc` looks at the `scr` attribute on `script` tags, it doesn't look at network request. In this case it should match `v1`.

AbhinavS96:
That's interesting. Because the chrome extension is picking it up as expected. The CLI is not working though.

Should I just change it to pick up v1?

AliasIO:
It should pick up `v1`, thanks!

AliasIO:
Thank you!

PR #5244, opened 2021-11-17
Add Shopaholic technology

FlusherDock1:
Adding https://shopaholic.one/
This is extension for October CMS


src/technologies/s.json
```diff
@@ -1468,6 +1468,28 @@
...
     "website": "https://www.shopgold.pl"
   },
+  "Shopaholic": {
+    "cats": [
+      6
+    ],
+    "cookies": {
+      "shopaholic_cart_id": ""
+    },
+    "dom": {
```
nurbek91, line 1478: Hey mate, thanks for PR . Let's remove the` dom.`
And instead of it we will write:
```
"js": {
      "ShopaholicCart": ""
    },
```
example: https://babylon64.ru/

src/technologies/s.json
```diff
@@ -1468,6 +1468,28 @@
...
+      "[class*='_shopaholic']": {
+        "exists": ""
+      }
+    },
+    "implies": [
```
nurbek91, line 1483: \\;confidence:100" I think this is not needed here

nurbek91 (changes requested):

src/technologies/s.json
FlusherDock1: Why?

src/technologies/s.json
FlusherDock1: I don' understand how your js section works.

nurbek91: image

src/technologies/s.json
nurbek91: "requires": "October CMS", 
"implies": "October CMS\\;confidence:99", 
It doesn't look logical. I guess I'm finding fault. If there are no mistakes, you can leave

AliasIO:
I fixed it up, thank you!

PR #5260, opened 2021-11-19
Add Nestify / hosting

nurbek91:
### website:
https://nestify.io/
### examples:
https://www.1hindi.com/
https://simplypure.com/
https://www.sexkontakte.com/


nurbek91:
Support:
use Cloudflare nameserver and all are different according to sites.

Not confirmed by support :
"headers": {
      "x-nestify-cache": ""
    },
The accounts team is not on the desk right now, we will inform you shortly.

PR #5265, opened 2021-11-19
Fix Invalid CSS selector pattern

basnetsoyuj:
previously: `link[*='/wp-content/themes/hueman/']`

Now: `link[href*='/wp-content/themes/hueman/']`

When I was processing this pattern in another place, I was gettting this error:

image


AliasIO:
Thank you!

PR #5304, opened 2021-11-23
fix(npm): skip external domains redirections

revolunet:
When we crawl a website, and some link points to an internal url (ex: `/a-propos`)
If that URL has a server-side redirection to some external domain, then the external url scan is also included in the results.

There was already some hostname filtering on the `links` parts but i think we should also check the hostname when a 30X error occurs.

Here's a brutal approach that prevent external urls but this produce an error in the log : `Navigation failed because browser has disconnected!`, not sure if this is the best approach.


AliasIO:
I think it should follow an initial redirect but perhaps not subsequent redirects when crawling recursively (i.e. all results should be from the same domain, even if the first requested redirected to a different domain). Would you agree?

revolunet:
I think if we're redirecting to a 3rd-party domain, it should not be included in the results because it will mix a different set of results

AliasIO:
This has been added in 1097edde542d06cd4d83bddb73f4d53fe4af8b15. By default, cross-domain redirects are only allowed on the first request. To disable redirects on all requests you can set `noRedirect` to `true` or pass `--no-redirect` or `-N`.

revolunet:
Awesome, thank you @AliasIO 💙

PR #5347, opened 2021-11-30
fix(driver.js): try catch status on analyse

MatthieuTartiere:
Error OnLine 998 of Driver.js cause cli.js crash.


AliasIO:
Please use Node.js 14 or higher.

PR #5378, opened 2021-12-06
Duplicate Squarespace template version tagging for "Squarespace Commerce" websites

tomccabe:
Adding some fields to Squarespace Commerce that were previously assumed to be inherited from Squarespace.


AliasIO:
Thanks!

PR #5407, opened 2021-12-12
fix: removing implies for stitches

mzaien:
removing implies from stitches because it works with any js/ts framework


AliasIO:
Thanks!

PR #5419, opened 2021-12-14
chore: add missing categories to Next.js

Atinux:
They are not listed in https://www.wappalyzer.com/technologies/static-site-generator/


AliasIO:
Thanks! The aim is to have one or two main categories, let me know which ones you think we should use.

PR #5457, opened 2021-12-28
Improve Gunicorn

baseplate-admin:
Gunicorn Implies Python


AliasIO:
Does Gunicorn exclusively work with Python applications or is Gunicorn itself written in Python?

baseplate-admin:
Gunicorn is written in python. And it only works with python wsgi applications. 

https://github.com/benoitc/gunicorn

AliasIO:
In the case of web servers, Python should only be implied if it serves an application written in Python. We use to imply Java for TomCat but it's confusing as it might be serving an application written in PHP. So unless the servers exclusively work with Python applications, the implies should be removed.

baseplate-admin:
But Gunicorn works exclusively with python.

AliasIO:
Ok great. :) What about the others you submitted?

baseplate-admin:
Bout others. Yea they are written in python. I hyperlinked Their Github page. You can take a look for yourself

AliasIO:
I mean, do they exclusively serve applications written in Python?

baseplate-admin:
Yep.

PR #5475, opened 2022-01-01
patch cgit

baseplate-admin:
Cgit should imply C.

Proof : 
- https://github.com/metajack/cgit


baseplate-admin:
This should work.

Because i added support [here](https://github.com/AliasIO/wappalyzer/pull/5476)

PR #5477, opened 2022-01-01
Patch gogs

baseplate-admin:
Gogs should imply Golang, Macaron.


baseplate-admin:
Again shouldn't fail because of [this](https://github.com/AliasIO/wappalyzer/pull/5478).

@AliasIO  please check

PR #5490, opened 2022-01-07
Update anime.js.png

baseplate-admin:
### Website : https://animejs.com

So the icon looks very weird in wappalyzer menu.

Capture

The example logo looks better than the current implementation.

image

So i essentially ripped the image from here and did some image magic :P


baseplate-admin:
Also on another note @AliasIO can we turn the style in the box to something purple-ish ? The current white box looks kinda weird( to me at least )

AliasIO:
Can you find an SVG version of the icon?

We could maybe use a light purple for the background. Many icons don't look good on a dark background.

baseplate-admin:
> Can you find an SVG version of the icon? 

Sure i will take a look tomorrow.

> We could maybe use a light purple for the background. Many icons don't look good on a dark background.

Enable dark background experimentally. Then slowly remove background from logos that doesn't have transparent background.

AliasIO:
The problem is with dark icons with a transparent background, like WordPress:

https://github.com/AliasIO/wappalyzer/blob/master/src/drivers/webextension/images/icons/WordPress.svg

baseplate-admin:
Maybe we can use this icon for wordpress instead?

AliasIO:
In this case yes but we can't always change a brand's colour. SVG files do support light and dark themes but it will be a lot of work to check and update 3,000+ icons.

baseplate-admin:
> Can you find an SVG version of the icon?

Found it.

> In this case yes but we can't always change a brand's colour. SVG files do support light and dark themes but it will be a lot of work to check and update 3,000+ icons.

I see your point. But in the end ``We could maybe use a light purple for the background.`` looks a good option to me.

baseplate-admin:
@AliasIO  what am i missing? Why is the check failing ?

I just merged your master into mine, so why is it failing?

AliasIO:
Thanks!

PR #5492, opened 2022-01-08
Fixed i30con detection

DevineDecrypter:
Thanks for reviewing my code @AliasIO , but the changes you made to the i30con object somehow disabled wappalyzer from picking up the technology in our website.
 With all due respect, I reverted the changes you made to make it work again.


AliasIO:
The `html` field has been deprecated. Can you share some examples of websites that use i30con so I can have a look?

DevineDecrypter:
Here`s the website https://30nama.com/ .
Some of the icons have the class **i30con-<something>** on them.

image

AliasIO:
Ok, so it should be `[class^='i30con']` instead of `[class$='i30con']`. Thanks!

PR #5518, opened 2022-01-17
Replace bulma.png with bulma.svg

baseplate-admin:
So bulma recently added a official SVG logo.
Heres the [issue](https://github.com/jgthms/bulma/issues/3461)


AliasIO:
Thanks! The SVG needs to be square but I'll fix that up.

PR #5553, opened 2022-01-28
Fix LaunchDarkly xhr definition

myfluxi:
* Uses array instead of string (as per definition)
* Breaks strict type checking


AliasIO:
What error did you get? All patterns can be either a string or array.

myfluxi:
See: https://github.com/AliasIO/wappalyzer/blob/master/README.md?plain=1#L390

Good news is, all other json values of all definitions are spot on (as per definition).

AliasIO:
Ah, good catch. I've updated the README. Also see the JSON schema:

https://github.com/AliasIO/wappalyzer/blob/master/schema.json

myfluxi:
Addressed in: https://github.com/AliasIO/wappalyzer/commit/21cef7d268a0a92efc2aa298480bdde29ac9ce65
and: https://github.com/AliasIO/wappalyzer/blob/master/schema.json

Thanks 👌🏼

PR #5556, opened 2022-02-01
Added Hyva as technology

jesse-deboer:
Hyva has become a well know new front-end of Magento 2, you can find sites using it here: https://hyva.io/showcases

The technology is way different from a normal Magento front-end and Wappalyzer does not always recognize this, for example this is https://www.limburgiavlaai.nl/ :
image


AliasIO:
Thanks!

PR #5564, opened 2022-02-02
update Layer0 definition to include response headers

tmountjr:
Layer0 sites include response headers in the form of `x-0-*` (https://docs.layer0.co/guides/response_headers) which also include the version of Layer0 used to deploy the site in the `x-0-version` header.

### DIFF
--- a/src/technologies/l.json
+++ b/src/technologies/l.json
@@ -185,6 +185,11 @@
       "layer0_eid": ""
     },
     "description": "Layer0 is an all-in-one Jamstack platform to develop, deploy, preview, split test, and monitor the company's frontend.",
+    "headers": {
+      "x-0-status": "",
+      "x-0-t": "",
+      "x-0-version": ""
+    },
     "icon": "Layer0.svg",
     "js": {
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

tmountjr:
@AliasIO just double-checking on the version string portion - looks like it was updated from `([\\d\\.]+)` to `([\\d.]+)`. The version string from Layer0 looks like this:

```
$ curl -I https://www.example.com
HTTP/2 200 
...
x-0-version: 3159 4.8.3 107 2022-01-27T15:27:54.948Z 1.2.0
```

The change removes the literal `.` from the capture group and while a period in the version string (`4.8.3` in this case) would be caught by the regex, it is a bit less specific. I suppose it would still work to leave it as-is, just wanted to check. Thanks!

AliasIO:
The period doesn't need to be escaped in a character class, it always matches a literal `.`. There's also no need to match to the end (`.+$`) so I removed that too.

tmountjr:
Ah, right, brain fart on the square brackets. Good to go then.

PR #5583, opened 2022-02-06
Fix Drupal Multisite - Expand asset source detection

kyletaylored:
Multisite detection was failing on known multisites, expanded detection to other assets.


AliasIO:
Thanks!

PR #5598, opened 2022-02-10
Update Blitz / performance

nurbek91:
###  website:
https://putyourlightson.com/plugins/blitz
https://plugins.craftcms.com/blitz
https://github.com/putyourlightson/craft-blitz
### examples:
https://imagecomics.com/
https://www.transparency.org/en/
https://thebreakthrough.org/
https://www.impraise.com/

### DIFF
--- a/src/technologies/a.json
+++ b/src/technologies/a.json
@@ -2544,7 +2544,7 @@
     "cats": [
       25
     ],
-    "description": "AntV G6 is a graph visualization framework in JavaScript.",
+    "description": "AntV G6 is a graph visualisation framework in JavaScript.",
     "icon": "AntV.svg",
     "js": {
       "g6.version": "([\\d\\.]+)\\;version:\\1",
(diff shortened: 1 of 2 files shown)


nurbek91:
Hey, If there is a comment everywhere:
"html": "<!-- Cached by Blitz on",

But for example, 80% have js, or header
headers: {
       "X-Powered-By": "^Blitz$"
     },
     "js": {
       Blitz: "",
       "blitzReplace": ""
     },

How will the Wappalyzer work after it first sees the header or js with 100% confidence, then the html check will not start?

AliasIO:
@nurbek91 All the checks will run even if it's already at 100% confidence. This is mainly because some of the other patterns may be needed to find a version number.

nurbek91:
> @nurbek91 All the checks will run even if it's already at 100% confidence. This is mainly because some of the other patterns may be needed to find a version number.

ah, ok thanks!

PR #5603, opened 2022-02-10
update Layer0 to read as PaaS as well as Performance

tmountjr:
Layer0 can function as both a platform (in the case of completely-hosted jamstack applications) and/or a performance enhancer (eg. a non-Node.js-based site being fronted by Layer0's CDN). Updated the category listing to include Layer0 as a PaaS platform rather than just a performance optimization.


AliasIO:
Thanks!

PR #5607, opened 2022-02-13
Sveltekit patch

baseplate-admin:
Sites with SvelteKit have a id = 'svelte-announcer' in dom.

image

Example : 
* https://svelte.dev/


AliasIO:
Thanks! I corrected the syntax and a formatting error.

PR #5611, opened 2022-02-14
Change ElasticSuite regex

fabianhafner:
The current `[\\S\\s.*]*` groups in ElasticSuite's second html regex match the entire html document (all non-whitespace, whitespace, literal . and *) and basically full text search Smile_Elasticsuite on the entire page. I don't think that was intended, since the pattern is surrounded by < >. Instead, I extended the first pattern from just input tags to all html tags.

### DIFF
--- a/src/technologies/e.json
+++ b/src/technologies/e.json
@@ -290,8 +290,7 @@
     },
     "description": "ElasticSuite is a merchandising suite for Magento and OroCommerce.",
     "html": [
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks! The `html` field is deprecated (in favour of `dom`) and the pattern seems to be covered by `scriptSrc` anyway, so I'll just remove it.

PR #5617, opened 2022-02-17
Fix Drupal Multisite

kyletaylored:
(The last PR wasn't quite finished, this will fix what was merged).

- Remove extraneous DOM detections
- Fix srcScript regex
- Update Acquia Site Factory (implies Drupal Multisite)


AliasIO:
Thanks!

PR #5637, opened 2022-02-25
Move django_language cookie detection from Django CMS to Django

thibaudcolas:
Spotted because I was reviewing two sites (https://vb-tpb.ch/fr/voyager/arrets/, https://sla.gs1.events/fr/) I had been told were built with Wagtail, but Wappalyzer reports them as both Wagtail and Django CMS (which would technically be possible, but very odd).

Looking at Wappalyzer’s checks, this is because of the `django_language` cookie detection. `django_language` is the default cookie name (see [Django docs](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-LANGUAGE_COOKIE_NAME) for it) for all Django projects that use Django’s translation support, it’s not specific to Django CMS.


AliasIO:
Thanks!

Issue #5635, opened 2022-02-25
Astro is not detected


sklar:
**Platform**
Which platform does the bug apply to?
* WebExtension (Google Chrome)

**Describe the bug**
Technology is not detected by the browser extension.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to [Astro website](https://astro.build/)
2. Click on Wappalyzer button
3. Technology is not detected.

<img width="1017" alt="Screenshot 2022-02-25 at 10 15 34" src="https://user-images.githubusercontent.com/2298407/155741158-c4a6aaa7-3d05-4530-92fa-f6a1f83c130d.png">

**Expected behaviour**
Technology is detected.

**Additional context**

Astro was introduced in [v6.9.14](https://github.com/AliasIO/wappalyzer/releases/tag/6.9.14) by #5504.
If I understand it correctly, the detection is implemented by the `href` attribute pattern.

I dare to recommend seconding it with the `class` attribute pattern as it was in the original new technology request.
```
astro-[\d\w]{8,}
```

nurbek91:
Hey @sklar updated, thanks!!!

PR #5641, opened 2022-02-28
Update Astro / page builder #5635

nurbek91:
### website:
https://astro.build/
### examples:
https://astro.build/

### DIFF
--- a/src/technologies/a.json
+++ b/src/technologies/a.json
@@ -3199,7 +3199,7 @@
           "href": "/_astro/(?:common|index)-[\\d\\w]+\\.css"
         }
       },
-      "span,section": {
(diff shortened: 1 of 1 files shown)


src/technologies/a.json
```diff
@@ -3198,6 +3198,11 @@
         "attributes": {
           "href": "/_astro/(?:common|index)-[\\d\\w]+\\.css"
         }
+      },
+      "span,section": {
```
sklar, line 3202: IMO the Astro hashed `class` could be assigned to **any dom element**, not just `span` or `section` 🤔

AliasIO: Updated!

PR #5687, opened 2022-03-09
Update SVG / PNG conversion script

kyletaylored:
Mostly this was frustrating me with the setMaxListener output. I believe this is mostly under the hood with the conversion package, so I adjusted and masked where I could. Ultimately, it's a much cleaner output.

- Attempt to increase listeners
- Remove excessive warning output
- Add better CLI output

convert-png-update

### DIFF
--- a/bin/convert.js
+++ b/bin/convert.js
@@ -1,6 +1,11 @@
 const fs = require('fs')
 const path = require('path')
 const { createConverter } = require('convert-svg-to-png')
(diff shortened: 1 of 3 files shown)


AliasIO:
Looks like there's an issue with running the converter in parallel, it saves the same image to different outputs. I changed it so it just does one a time for now (a2577205b996d4c208eb65cb3ca0e8369ca6b880).

Issue #5844, opened 2022-03-21
Track 'Qwik'


rockeynebhwani:
Vendor site - https://github.com/builderio/qwik
Category - Web framewoks
Example Site - https://www.builder.io/
Detection - Look for 'q:version' attribute in HTML tag.

steve8708:
hey @rockeynebhwani - we are making an update to Qwik now to support microfrontends. This means that q:version may not always be on the HTML tag, and instead can be on any element. So if we change this to instead just look for `document. querySelector('[q\\:version]')` we can catch that use case too

cc @manucorporat @mhevery

PR #5848, opened 2022-03-21
Add Qwik / web framework #5844

nurbek91:
### website:
https://github.com/builderio/qwik
### examples:
https://www.builder.io/

### DIFF
--- a/src/technologies/q.json
+++ b/src/technologies/q.json
@@ -237,5 +237,27 @@
     },
     "scriptSrc": "\\.quora\\.com/",
     "website": "https://quoraadsupport.zendesk.com/hc/en-us/categories/115001573928-Pixels-Tracking"
+  },
+  "Qwik": {
+    "cats": [
+      18
+    ],
+    "description": "Qwik is designed for the fastest possible page load time, by delivering pure HTML with near 0 JavaScript.",
+    "icon": "Qwik.svg",
+    "dom": {
+      "html": {
+        "attributes": {
+          "q:version": "^([\\d\\.]+(?:-\\d+)?)\\;version:\\1"
+        }
+      },
+      "div": {
+        "attributes": {
+          "q:obj": "\\;confidence:50",
+          "q:host": "\\;confidence:50"
+        }
+      }
+    },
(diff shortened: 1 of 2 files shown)


src/technologies/q.json
```diff
@@ -237,5 +237,27 @@
...
+      },
+      "div": {
+        "attributes": {
+          "q:obj": "\\;confidence:50",
+          "q:host": "\\;confidence:50"
+        }
+      }
+    },
+    "oss": true,
+    "website": "https://github.com/builderio/qwik"
```
mhevery, line 261:
```suggestion
    "website": "https://qwik.builder.io"
```

I know it is not public just yet, but should be soon.

src/technologies/q.json
```diff
@@ -237,5 +237,27 @@
...
+  "Qwik": {
+    "cats": [
+      18
+    ],
+    "description": "Qwik is designed for the fastest possible page load time, by delivering pure HTML with near 0 JavaScript.",
+    "icon": "Qwik.svg",
+    "dom": {
+      "html": {
+        "attributes": {
+          "q:version": "^([\\d\\.]+(?:-\\d+)?)\\;version:\\1"
```
mhevery, line 250: The `q:version` can be on any element not just on `html` Is there a way to express it?

src/technologies/q.json
```diff
@@ -237,5 +237,27 @@
...
+        }
+      },
+      "div": {
+        "attributes": {
+          "q:obj": "\\;confidence:50",
```
mhevery, line 255: It is possible that there is no `q:obj` if the content is static.
There should always be at least on `q:host` attribute so I think that is fine. But why not just rely on `q:version`?

nurbek91: Hey @mhevery  , added **body** tag.
Versions should now be checked for all these tags **html,body,div**
But then my knowledge ends there.
Further @AliasIO , i need your help mate.

AliasIO: I've updated it to always just use `q:version`, assuming it's always present (on any element).

PR #5958, opened 2022-03-30
Add Cart.js / js lib #5957

nurbek91:
### website:
https://cartjs.org/
### examples:
**type** https://cdn.shopify.com/s/files/1/1279/0445/t/18/assets/rivets-cart.min.js
https://www.justeliquids.com/
https://elegantautoretail.com/
**type** https://cdnjs.cloudflare.com/ajax/libs/shopify-cartjs/0.4.1/rivets-cart.min.js
https://www.evolvevapors.com/
https://tonymoly.us/
https://www.aeraforhome.com/


rockeynebhwani (commented):
@nurbek91 - Are you looking for both 'cart.min.js' and 'rivets-cart.min.js' in regex or just one? Sorry.. my regex is really bad..

nurbek91:
> @nurbek91 - Are you looking for both 'cart.min.js' and 'rivets-cart.min.js' in regex or just one? Sorry.. my regex is really bad..
@rockeynebhwani Hey mate. If I'm not wrong about anything. Yes, we are looking at both options. Where is the version we take the version.

nurbek91:
> @nurbek91 - Are you looking for both 'cart.min.js' and 'rivets-cart.min.js' in regex or just one? Sorry.. my regex is really bad..

I also checked the internals of the file. It seems to confirm that this is one and the same technology.

rockeynebhwani:
In that case, happy to merge

PR #6028, opened 2022-04-05
Improve Froala detection

jkmartindale:
- `dom` rule replaces `html` rule that inadvertently expected discriminating class name to appear last in the class list
- Version detection for Froala 2+
- Additional discriminating class for popup editors (existing code doesn't recognize Froala on [these](https://froala.com/wysiwyg-editor/examples/edit-in-popup/) official [demos](https://froala.com/wysiwyg-editor/examples/init-on-button/))
- Additional discriminating class for Froala 1.x

### DIFF
--- a/src/technologies/f.json
+++ b/src/technologies/f.json
@@ -1380,12 +1380,15 @@
       24
     ],
     "description": "Froala Editor is a WYSIWYG HTML Editor written in Javascript that enables rich text editing capabilities for applications.",
-    "html": "<[^>]+class=\"[^\"]*(?:fr-view|fr-box)",
+    "dom": ".fr-view, .fr-box, .fr-popup, .froala-box",
     "icon": "Froala.svg",
     "implies": [
       "jQuery",
       "Font Awesome"
     ],
+    "js": {
+      "FroalaEditor.VERSION": "([\\d\\.]+)\\;version:\\1"
+    },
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks! You could create two separate `dom` rules and tag them with the version number.

jkmartindale:
Do `dom` rules support version tagging? I found that this worked fine:
```json
"dom": [
  ".fr-view,.fr-box,.fr-popup",
  ".froala-box"
],
```

But as soon as I added the version tags, DOM-based detection broke entirely:

```json
"dom": [
  ".fr-view,.fr-box,.fr-popup\\;version:2+",
  ".froala-box\\;version:1"
],
```

I'm not at all familiar with the Wappalyzer driver code but from a cursory glance I'm wildly speculating `dom` rules might get passed directly to `querySelectorAll()` without checking for version or confidence:

https://github.com/wappalyzer/wappalyzer/blob/f56c6fdf905be8ef870001002762111e7cbaea5e/src/drivers/webextension/js/dom.js#L21-L28

AliasIO:
It works, just not in the shorthand notation. You can do this:

```
"dom": {
  ".froala-box": {
    "exists": "\\;version:1"
  },
  ".fr-view, .fr-box, .fr-popup": {
    "exists": "\\;version:2"
  }
}
```

PR #6037, opened 2022-04-06
Updated categorisations and removed 'Fast Checkout'

rockeynebhwani:
Fast checkout is shutting down and hence removing - https://www.fast.co/


AliasIO:
Thanks!

PR #6043, opened 2022-04-07
Update Salla / ecommerce

nurbek91:
just prettified json


AliasIO:
You can run `yarn run prettify` to fix up all the technology files at once. :)

PR #6106, opened 2022-04-16
Update Express, Phusion Passenger / web server

nurbek91:
### examples
https://yaroslavhrabovets.in.ua/
https://ibertronica.es/
https://www.domena.cz/

Hey @AliasIO , my knowledge is not enough for web servers. I have a question
In all these examples there is such a trace
`x-powered-by: Express, Phusion Passenger(R) 6.0.12`

Express **implies** Node.js , 
Can there be multiple web servers on one site?

Apache,
Express,
Node.js.
Phusion Passenger

or 

Ngnix,
Express,
Node.js.
Phusion Passenger


AliasIO:
It's possible, usually there's one but sometimes two (e.g. a proxy server).

Issue #5912, opened 2022-03-26
Improve categorisation of apps


rockeynebhwani:
In last few months, we have added bunch of categories which allows us to better categorise tech listed under 'Widgets' / 'Miscellaneous'. Opening this issue to identify such apps and update the categories wherever applicable.

Issue #6061, opened 2022-04-08
Improve 'Shopify' detection


rockeynebhwani:
Following shopify sites where detection is not working because these sites are using headless architecture. You may see it being detected as you progress through the checkout flow - 

- https://iliabeauty.com/ (Check Detection 1)
- https://kotn.com/
- https://www.seedlipdrinks.com/ (Check Detection 2)
- https://nakedlabs.com/ (Check Detection 3)
- https://ensorings.com/ (Check Detection 1)
- https://www.analuisa.com/
- https://www.onequince.com/
- https://www.tamaramellon.com/products/astral-suede-sandals?color=Teal%20Yade&origin=hp-carousel

Will add more examples and provide some directions on how we can improve this

Detection - 

1) Look for '?shop=*.myshopify.com' within SCRIPT SRC URL. This is typically included by loading Shopify APP JS and this can be a useful signal 
2) Refer https://github.com/wappalyzer/wappalyzer/issues/6048. We can 'Imply' Shopify when 'Shopify Buy' button is found
3) Look for URLs from domain '//cdn.shopify.com/' in `<video>` tag 'poster' attribute  (may be give this confidence level of 50%)
4) Also see https://github.com/wappalyzer/wappalyzer/issues/6223

nurbek91:
.

nurbek91:
still required

nurbek91:
https://porijewelry.faire.com/product/p_f9827vvus7

rockeynebhwani:
.

PR #6123, opened 2022-04-17
More updates to categorisation

rockeynebhwani:
https://github.com/wappalyzer/wappalyzer/issues/5912

### DIFF
--- a/src/technologies/h.json
+++ b/src/technologies/h.json
@@ -1033,7 +1033,8 @@
   },
   "HulkApps Form Builder": {
     "cats": [
-      73
+      73,
+      100
     ],
     "description": "HulkApps Form Builder is an application that creates customizable, job-specific forms for unit needs.",
     "icon": "HulkApps.svg",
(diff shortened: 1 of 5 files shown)


nurbek91:
Hey @rockeynebhwani ,thanks!!!
How do you look if  replace for 5  HulkApps apps 
`"implies": "Shopify",`
to 
`"requires": "Shopify",`

rockeynebhwani:
@nurbek91  - There is a growing trend of sites taking headless approach which means that it's not always possible to detect 'Shopify' but it's still possible to detect these apps.. and in that case, if we use 'implies', it helps with detection of 'Shopify' platform. Please also see - https://github.com/wappalyzer/wappalyzer/issues/6061

nurbek91:
Hey @rockeynebhwani  Thank you, I didn't consider this question from this side.
I previously added the requires attribute for all Shopify applications
Thought it would help optimize performance.

Hey @AliasIO  What can you advise?
Change for all Shopify apps requires to implies?

AliasIO:
Yes, it's a good point. Let's use `implies` for Shopify.

PR #6151, opened 2022-04-21
Add Acquia Cloud Platform CDN / Update Fastly

kyletaylored:
Acquia Cloud Platform CDN runs on Fastly, so this will add the new product, and update the Fastly detection that is more agnostic.


AliasIO:
Thanks!

PR #6206, opened 2022-04-26
Update The Hut Group / Miscellaneous

nurbek91:
### website
https://www.thg.com/
### examples
https://www.wappalyzer.com/technologies/ecommerce/the-hut-group?utm_source=popup&utm_medium=extension&utm_campaign=wappalyzer

### DIFF
--- a/src/technologies/t.json
+++ b/src/technologies/t.json
@@ -701,13 +701,16 @@
   },
   "The Hut Group": {
     "cats": [
-      6
+      19
     ],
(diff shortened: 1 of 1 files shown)


rockeynebhwani:
@nurbek91 / @AliasIO  - Why did we move this from 'Ecommerce' to 'Miscellaneous'. Hut group has a platform called 'Ingenuity' ('https://www.thgingenuity.com/) which they use for some of their own brands but also provide this platform as a service for other brands... I think this should stay under 'Ecommece'. You may want to rename the platform to 'Ingenuity' if you want.

PR #6213, opened 2022-04-27
Update Shopify / ecommerce

nurbek91:
### example
https://www.cutsclothing.com/


nurbek91:
> @nurbek91 - Did you see this ? #6061

Thanks mate, I’ll have a look

PR #6246, opened 2022-05-04
Add The Hotels Network / personalisation, analytics

nurbek91:
### website
https://thehotelsnetwork.com/en/pricing
### examples
https://www.temptation-experience.com/
https://www.camelbackresort.com/
http://burnhamsuitesbaguio.com/
https://www.oakshotels.com/en/

` "content-security-policy-report-only": "\\.thehotelsnetwork\\.com"` is it ok ?
https://www.anantara.com/en/

locally does not take the version
possibly a bug if the version is nested more than two levels
` "thn.data.version": "([\\d\\.]+)\\;version:\\1"`


AliasIO:
Both should work.

PR #6288, opened 2022-05-17
Add Strattic

jdevalk:
Adds support for detecting Strattic, static WordPress hosting.

strattic-wappalyzer


AliasIO:
Thanks!

jdevalk:
Hey @AliasIO I noticed you actually made it require `WordPress` when, since Strattic only hosts WordPress sites, it actually _implies_ it. Since it's an HTTP header detection it might actually be faster that way.

AliasIO:
Ok, I'll change it back.

PR #6290, opened 2022-05-17
Add Yoast SEO for Shopify

jdevalk:
* Adds Yoast SEO for Shopify detection.
* Also change Yoast SEO for WordPress from requiring WordPress to implying it, as its detection is fairly "certain" and improves WordPress detection. For example see https://www.vanderakt.nl/ 

CleanShot 2022-05-17 at 14 14 12@2x


AliasIO:
Thanks! Would you be able to reduce the icon dimensions to 32x32?

jdevalk:
> Thanks! Would you be able to reduce the icon dimensions to 32x32?

Done!

AliasIO:
Thanks!

PR #6293, opened 2022-05-18
Fix the problem that missing slash cannot be copied during docker build

ajinwu:
Adding multiple files to a directory in docker requires a slash, if there is no slash then it cannot be copied to docker properly, this fix can build docker images properly


AliasIO:
Thanks!

PR #6297, opened 2022-05-19
Add Reactive detection

reactiveonline:
This commit adds Reactive platform to the list of available technologies at wappalyzer.


AliasIO:
Thanks!

PR #6298, opened 2022-05-21
Replace png to svg and other fixes

nurbek91:
update icons and add pricings


rockeynebhwani:
Thanks @nurbek91 for sorting out my mess :-)

PR #6307, opened 2022-05-25
Improve Froala 1.x detection

jkmartindale:
Discussion at #6028 had valid DOM rules to discriminate Froala 1.x from 2+ but I guess I forgot to get that merged


AliasIO:
Thank you!

PR #6398, opened 2022-06-17
Add Tebex / ecommerce

nurbek91:
### website
https://www.tebex.io/
### examples
https://store.hypixel.net/
https://store.nopixel.net/
https://store.rustacademy.net/


nurbek91:
### Validation failed becuase "Sass" added here #6386

### Merged interesting points with  **confidence** parameter

### Summation doesn't work here shows 33 %
image

### Summation  work here shows correct number
image

### Confidence issue in extension
https://buzones.solobuzones.es/
image
3 tech with 25% confidence which doesn't show
image

AliasIO:
Thanks, I'll have a look at these.

AliasIO:
@nurbek91 Technologies are hidden in the extension when `confidence` is lower than 50. Tebex doesn't show up on the example sites because it has `"requires": "Cart Functionality"`.

PR #6402, opened 2022-06-19
Update React / js framworks

nurbek91:
### examples
https://cosbots.imonitor.live/register
https://hh.ru/
https://www.linkedin.com/feed/
image

### DIFF
--- a/src/technologies/r.json
+++ b/src/technologies/r.json
@@ -421,7 +421,7 @@
           "_reactRootContainer": ""
         }
       },
-      "div[id*='react']": {
+      "div[id*='react-root']": {
         "attributes": {
           "id": "react-root"
         }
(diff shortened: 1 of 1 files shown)


src/technologies/r.json
```diff
@@ -420,7 +420,17 @@
         "properties": {
           "_reactRootContainer": ""
         }
-      }
+      },
+      "div[id*='react']": {
+        "attributes": {
+          "id": "react-root"
```
AliasIO, line 426: Should this be `div[id='react-root']` or  `"id": "react"`? Just matching `react` is probably too broad.

nurbek91: Updated like this. I hope correctly
` "div[id*='react-root']":`

AliasIO: In that case you can do this:

```
"div[id*='react-root']": {
  "exists": ""
}
```

PR #6418, opened 2022-06-24
Edgio name change

tmountjr:
Layer0 by Limelight has rebranded as Edgio. This commit removes Layer0 from the technology list and replaces it with Edgio (the detection is the same for now). Also updated the icon in use but the gradient is a bit tricky and I used inkscape's mesh gradient to get it right. If that's an issue (it's got a JS polyfill embedded) I've got the PNG we can use instead.


tmountjr:
The check failed for me locally because there's a missing SVG for a different technology (not mine - it's for nopStation).

nurbek91 (approved):
Hey mate , thanks for PR!!!

AliasIO:
Thanks!

Issue #5972, opened 2022-03-30
Track 'Lazysizes unveilhooks plugin'


rockeynebhwani:
Vendor site - https://github.com/aFarkas/lazysizes/tree/gh-pages/plugins/unveilhooks
Category - Javascript Libraries / Performance
Example Site - https://www.missouriquiltco.com/products/creative-grids-cutting-mat-24-x-36
Detection - Look for script 'ls.unveilhooks.js' or ''ls.unveilhooks.min.js

PR #6452, opened 2022-06-30
Fix lazysize unveilhooks regex

mlec1:
According to the ticket https://github.com/wappalyzer/wappalyzer/issues/5972, we are looking for script 'ls.unveilhooks.js' or 'ls.unveilhooks.min.js'.

It is just a small fix in the regex in order to escape the point "."


nurbek91 (approved):

AliasIO:
Nice catch, thanks!

PR #6479, opened 2022-07-04
Update Chakra UI dom Detection

ysm0622:
No more `html[style*='chakra-ui-color-mode']` dom from Chakra UI v2.

examples
- https://upstash.com/
- https://chakra-ui.com/


AliasIO:
Thanks!

PR #6481, opened 2022-07-05
chore: Remove duplicate icon

kingthorin:
Having both upper and lower case variants on windows is problematic. The files are identical (SHA256 - ce7e91d4f840460983ddd05af067a0e19bb13df4114dfd4f3bf2bb24c1d1dc73)

I decided to keep the all lower variant.

- https://github.com/wappalyzer/wappalyzer/blob/c4704aa76175a98d4212bcaa126ceb1473e51e8e/src/drivers/webextension/images/icons/Deno.svg
- https://github.com/wappalyzer/wappalyzer/blob/a959eae80bf8f03bb047ce3119d85099cd199c95/src/drivers/webextension/images/icons/deno.svg


kingthorin:
@AliasIO do you think this could be added to the validation workflow somehow?

AliasIO:
I don't think so as I'm on Mac OS myself (case insensitive file system). The duplicate file would only appear on windows if it was renamed from upper to lower-case or vice versa.

PR #6514, opened 2022-07-20
Added description field to the output of CLI driver

kirillkovnat:
Description field exposed in the technologies JSONs, but not showed in the output. This PR fixes this issue


AliasIO:
Thanks! 👍

PR #6515, opened 2022-07-20
Fix URL property from request object

kyletaylored:
I've been running into a few issues related to detections not being picked up when the detection regex seems to be correct. After a bit debugging, I discovered that the URL being passed to the script detection was undefined (using an old convention, I'm assuming), which was causing the script detection to error out.

It looks like in Chrome, the `chrome.webrequest.oncompleted` sends back the request object which uses the url property instead of the current `documentURL` property (see link below).

https://developer.chrome.com/docs/extensions/reference/webRequest/#property-onCompleted-callback-details-url

The `request.url` prop is already being used, so I'm guessing this was just overlooked.

### DIFF
--- a/src/drivers/webextension/js/driver.js
+++ b/src/drivers/webextension/js/driver.js
@@ -439,9 +439,7 @@ const Driver = {
 
     const scripts = (await response.text()).slice(0, 500000)
 
-    Driver.onDetect(request.url, analyze({ scripts })).catch(
-      Driver.error
-    )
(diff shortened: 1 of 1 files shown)


AliasIO:
We need the document URL as the script can be loaded from a different domain. It looks like `request.documentURL` is Firefox only, the closest thing Chrome seems to have is `request.initiator`. Implemented in 90c5c6f41c9e0b8eb3370c6fd2149ddeabef81e4.

PR #6520, opened 2022-07-20
Add Acquia Cloud Next (and other AWS services)

kyletaylored:
Acquia Cloud released a second version of their platform built on top of Kubernetes and utilizing different AWS services based on their architecture overview: https://docs.acquia.com/cloud-platform/arch/

This will add Acquia Cloud Next and Acquia Cloud Classic with technologies separated by each, while also adding some new dependent AWS services.

- Add Acquia Cloud Next
- Add Acquia Cloud Classic
- Add Amazon Aurora
- Add Amazon EFS
- Add Red Hat Gluster

Other formatting updates were made based on `yarn build`.


AliasIO:
Thanks! I'll shrink the icons down to 32x32 px. 😉

kyletaylored:
Ahh, I keep forgetting the size!  😂 I'll remember next time

AliasIO:
Made [some changes](8d4e20a3bc374e8d22ddabcaa15dbe50a14aea9b) based on info from Acquia's CTO:

> `x-ah-environment`: prod is used by both Acquia Cloud Next and Acquia Cloud Classic.
> 
> When migrating customers from Classic to Next, we spin up a temporary test environment.  This temporary test environment is a clone of their site running on Acquia Cloud Next.  Those temporary test/migration environments set `x-ah-environment: next`. But, after they switch their test environment to production, the header changes to `x-ah-environment: prod`.
> 
> I checked for other HTTP headers that could help differentiate between Classic of Next but there aren't any: https://dri.es/headers?url=https://dri.es.  In other words, there is no reliable way to differentiate between the two of them, except maybe for testing IP address ranges. 
> 
> It is true that Acquia Cloud Next and Acquia Cloud Classic use different technologies.  Acquia Cloud Next uses AWS Aurora instead of Acquia managed MySQL instances on EC2.  Similarly, the file system moved from Gluster FS to AWS Elastic FS.
> 
> I'm not sure the distinction matters.  First, the names "Acquia Cloud Next" and "Acquia Cloud Classic" exist mainly for customer communication purposes.  The product is called "Acquia Cloud".  Second, we are migrating 100% of our customers to Acquia Cloud Next.  At this point, we are migrating hundreds of customers per month.  90% of our customers will be eligible to migrate by the end of 2022.  Once the migration is done, the names "Acquia Cloud Next" and "Acquia Cloud Classic" will disappear.  In short, the distinction is just temporary and for customer communication purposes only.

PR #6528, opened 2022-07-21
Adding new CPEs to various technologies

xmco:
Small MR to fix 1 erroneous CPE and add new CPE refs to various technologies.


AliasIO:
Thanks!

PR #6549, opened 2022-07-24
Fix exiting with error code when help is called

undergroundwires:
`--help` is an accepted option and should not return `1` (error) as exit code but success `0`. However, the CLI will still return `0` if help text is forced due to missing URL.


AliasIO:
Thanks!

Issue #6575, opened 2022-07-26
'utterances' not detected


John4266:
**Platform**
Which platform does the bug apply to?
* WebExtension (Google Chrome, Mozilla Firefox)

**Describe the bug**
Wappalyzer doesn't detect it

**To Reproduce**
go to site https://haxe.org/documentation/introduction/compiler-targets.html
check the extension

**Additional context**
Detection should work like this
Find `iframe` tag with class of `utterances-frame` and `src` that has `https://utteranc.es/utterances.html`

PR #6584, opened 2022-07-27
fix utterances pattern #6575

DavidKessel:
Updated the dom pattern for "Utterances" as suggested in issue #6575. This fixes the issue.
Cheers!


AliasIO:
Thanks!

PR #6589, opened 2022-07-29
Google App Engine: fix website, lower confidence in header

kevmoo:
"Server": "Google Frontend" could be a number of other things, including
Google Cloud Run

### DIFF
--- a/src/technologies/g.json
+++ b/src/technologies/g.json
@@ -1400,9 +1400,6 @@
     "cats": [
       22
     ],
-    "headers": {
-      "Server": "Google Frontend\\;confidence:30"
-    },
     "icon": "Google App Engine.svg",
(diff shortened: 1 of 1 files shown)


AliasIO:
The lower confidence isn't that useful unless there's a way for multiple patterns to add up to 100. Low confidence results are ignored so we'd never positively identify Google App Engine this way. We may just need to remove the header pattern if it's unreliable.

kevmoo:
I'd just drop it. There are no "certain" signal from App Engine. That means there'd be no reporting for App Engine – but that'd be better than a weak reporting.

The other option would be to put in `Google Cloud Run` at 50% or something

AliasIO:
I removed the pattern. I'll leave Google App Engine in there in case it can be implied by other technologies.

PR #6603, opened 2022-08-04
Add TripleWhale / Ecommerce

Overload119:
### website
https://triplewhale.com/

### examples
```
https://thehermoza.com/
https://freshk9.com/
https://www.glamrdip.com/
https://hoplark.com/
https://myobvi.com/
```


src/technologies/t.json
```diff
@@ -1,4 +1,13 @@
 {
+  "TripleWhale": {
+    "cats": [
+      10
+    ],
+    "html": "triplewhale-pixel.web.app\/triplepx.txt",
```
nurbek91, line 6: The html parameter is resource intensive and can be used when all other methods are not possible.
In your case, we can replace it with 
 "xhr": "triplewhale-pixel\\.web\\.app",

src/technologies/t.json
```diff
@@ -1,4 +1,13 @@
...
+    "saas": true,
+    "icon": "triplewhale.png",
```
nurbek91, line 8: Preferred icon format is svg with size 32x32
[Triple Whale.svg.zip](https://github.com/wappalyzer/wappalyzer/files/9264825/Triple.Whale.svg.zip)

src/technologies/t.json
```diff
@@ -1,4 +1,13 @@
 {
+  "TripleWhale": {
```
nurbek91, line 2: The correct name of technologies is two words written separately? Triple Whale

We can add description:

`"description": "Triple Whale is a mobile app used by ecommerce businesses to track profits, blended ROAS, LTV, and much more, integrating into a single easy-to-use mobile app.",`

pricing:
```
"pricing": [
      "mid",
      "recurring"
    ],
```

nurbek91 (changes requested):
Hey mate, thanks for PR!!
Needs an update to perfect your PR

src/technologies/t.json
Overload119: That description is not quite right - Triple Whale has an app, but it's not known as that.
I will pull a description from Crunchbase.

AliasIO:
Thanks!

PR #6617, opened 2022-08-06
Improve mailman CPE and implication

undergroundwires:
Remove Django implication from mailman. Mailman is not necessarily used
with Django but always with Python. Also add missing CPE.


AliasIO:
Thanks!

PR #6619, opened 2022-08-07
Add HSTS detection

InstabilSpielt:
The [Strict-Transport-Security](https://developer.mozilla.org/docs/Web/HTTP/Headers/strict-transport-security) header is used to determine whether the web server enforces HTTPS.


InstabilSpielt:
Sorry for the many commits for relatively little change, somehow I forgot a few things at the beginning. 😅

AliasIO:
Thanks!

PR #6655, opened 2022-08-16
Add Open-Xchange App Suite / Webmail

vantezzen:
This PR adds detection for the Open-Xchange App Suite webmail technology. Their SaaS software is used by GoDaddy Mail, Ionos (former 1&1), Rackspace, web.com mail and some Universities to host their webmailers.

Most companies (like GoDaddy) add their own, standalone login screen in front of the app suite themselves. This unfortunately results in the technology not being readable there, but the detection works on the app suite itself, i.e. after logging in.

# More information

https://www.open-xchange.com/
https://en.wikipedia.org/wiki/Open-Xchange
https://mail.ionos.com/
https://email.godaddy.com/


AliasIO:
Thank you!

PR #6672, opened 2022-08-26
Simplify Edgio name

tmountjr:
Product team requested that we simplify the name reported in Wappalyzer since "App Platform" is one distinct offering that Edgio provides, and isn't representative of the entire suite of tools.


AliasIO:
Thanks!

PR #6689, opened 2022-08-28
Update Google Analytics / analytics

nurbek91:
### example
https://www.cityclima.de/

image


tunetheweb:
I already see GA being detected on that site (after I accept the Cookie Banner), so not sure why this is necessary?

image

AliasIO:
I'll close this as the `gtag` object doesn't always imply GA.

PR #6709, opened 2022-08-30
Add FonServer / font-scripts

nurbek91:
### website 
https://fontserver.ir/
### examples
https://sam.behzisti.net/
http://ujiroft.ac.ir/
https://tasjil.urd.ac.ir/
https://khabarnews.com/
https://the-vv.com/login
https://veisa.ir/
https://javanexchange.com/
https://afaqtel.ir/css/custom.css
https://pinup-2022.ru/
### In one example, here is such a pattern. Is it worth it to take it through html? or is it possible via css @AliasIO 
https://justeducation.ir/
image


AliasIO:
You can use `css` for this. In this case you could do both as CSS is not always accessible depending on the website's configuration.

PR #6710, opened 2022-08-30
Add Skeleton / ui-frameworks

nurbek91:
### website 
http://getskeleton.com/
https://github.com/dhg/Skeleton Popular
### examples
https://mkvtoolnix.download/
http://ethcombo.com/
http://btcheat.com/

### made it to define only with standard paths and file names to avoid a lot of false positive results.
Also put 49 percent confidence if one file is only determined not to show in the extensions.
image


AliasIO:
The confidence should add up to `100`. The patterns do seem very generic, even if both files are there I'd expect some false positives.

nurbek91:
Hey mate, I don’t see any more opportunities, then we can close this request

Issue #6729, opened 2022-09-01
Over detection of Google Tag Manager


tunetheweb:
**Platform**
All

**Describe the bug**
Google Tag Manager detection is currently incorrect as it includes all Gtag implementations.

Gtag is a way of loading a number of Google resources including Google Analytics, Google Ad Words though I don’t think Google Tag Manager itself. It is hosted on www.googletagmanager.com and sets the `google_tag_manager` JavaScript variable.

Compare these:

- https://www.googletagmanager.com/gtag/js?id=UA-12345678-1 - this is Google Universal Analytics (identified by UA at the start of the id).
- https://www.googletagmanager.com/gtag/js?id=AW-123456789 - this is Google Ad Words (identified by AW at the start of the id.
- https://www.googletagmanager.com/gtm.js?id=GTM-ABC12DE - this is Google Tag Manager (identified by GTM at the start of the id and the use of gtm.js script).

Currently Wappalyzer classifies all as "Google Tag Manager" when only the last is.

Google tries to encourage Gtag with that as the given snippet for things like Google Analytics, even though those services CAN load separately, hence why it's used a lot. Note GTM itself cannot be installed via GTag (see above it's calling gtm.js not gtag.js).

GTag may share some of the infrastructure as Google Tag Manager, but it is not GTM as most users would see it (it cannot be used to add other tags).

**To Reproduce**
1. Check detections on https://almanac.httparchive.org/en/2021/
2. GTM is shown as used despite it not being used on this site

**Expected behavior**
GTM should be detected on sites that use GTM

**Additional context**

The current Google Tag Manager implementation is this:

https://github.com/wappalyzer/wappalyzer/blob/a1e310c2daa5ec7425febf13c69afb4706f4e55c/src/technologies/g.json#L1618-L1635

This includes "googletagmanager\\.com/gtag/js" which was added in #3398. I believe this is incorrect and the example site listed (https://iworkedon.com/) is using Google Analytics loaded via GTag not Google Tag Manager. I'm not sure if removing that again will be sufficient or if the html and js detections will still detect this as GTM?

Separately, we should probably add a new technology called Gtag or Google Tag to show those using Gtag to load GA, Adwords or whatever else.

nurbek91:
Hey there, thanks for request!!
#3398 This is already an old PR and the added piece of code is no longer in GTM
`"googletagmanager\\.com/gtag/js" `

[...]

PR #6731, opened 2022-09-02
Update Google Tag Manager / tag #6729

nurbek91:
#6729


tunetheweb:
I already see GA being detected on that site (after I accept the Cookie Banner), so not sure why this is necessary?

image

PR #6780, opened 2022-09-19
Improve Polylang detection

sebastienserre:
Add cookie + x-redirected-by header (value can be "Polylang" or "Polylang Pro")


AliasIO:
Thanks!

PR #6794, opened 2022-09-26
Added info for Inertia.js

jxxe:
- It's written Inertia.js, not Inertia like the logo
- Laravel or Ruby on Rails are implied by Inertia.js
- The confidence numbers come from the [Laravel](https://github.com/inertiajs/inertia-laravel/network/dependents) and [Rails](https://github.com/inertiajs/inertia-rails/network/dependents) repositories (16,438 vs 128 uses)
- It also implies React, Vue, or Svelte but I think those can be detected better other ways

### DIFF
--- a/src/technologies/i.json
+++ b/src/technologies/i.json
@@ -486,11 +486,11 @@
     },
     "website": "http://indyproject.org"
   },
-  "Inertia": {
+  "Inertia.js": {
     "cats": [
       12
     ],
-    "description": "Inertia is a protocol for creating monolithic single-page applications.",
+    "description": "Inertia.js is a protocol for creating monolithic single-page applications.",
     "dom": {
       "div[data-page*='component']": {
         "attributes": {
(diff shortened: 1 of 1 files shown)


src/technologies/i.json
```diff
@@ -501,6 +501,10 @@
     "headers": {
       "X-Inertia": ""
     },
+    "implies": [
+      "Laravel\\;confidence:99",
+      "Ruby on Rails\\;confidence:1"
```
AliasIO, line 506: Implies should not have `confidence`, unless Inertia.js always requires Laravel or RoR, these should be removed.

jxxe: Inertia.js is a tool that makes it easier to build frontends for Laravel and Ruby on Rails apps, so it always requires one of them.

AliasIO: The `implies` means it will always positively detect both 'Laravel' and 'Ruby on Rails' here. I'll remove these.

AliasIO:
Thanks!

PR #6827, opened 2022-10-10
Add Ripple

seanhamlin:
Example sites:

* https://www.vic.gov.au/
* https://www.coronavirus.vic.gov.au/


src/technologies/r.json
```diff
@@ -1593,6 +1593,23 @@
...
   },
+  "Ripple": {
+    "cats": [
+      12
+    ],
+    "description": "Ripple is the frontend framework for Single Digital Presence, delivered using Nuxt and Vue.js.",
+    "headers": {
+      "x-sdp-app-type": "ripple"
+    },
+    "icon": "ripple.png",
```
nurbek91, line 1604: 32x32
Ripple

seanhamlin: I was not aware that the icon has to be 32x32. Will attempt to find an SVG, or else will commit a smaller file.

PR #6830, opened 2022-10-10
Update ProcessWire detection.

seanhamlin:
Example site 

* https://www.antarctica.gov.au/


src/technologies/p.json
```diff
@@ -2504,17 +2504,21 @@
     "cats": [
       1
     ],
-    "description": "ProcessWire is an open source content management system (CMS) and framework (CMF).",
+    "description": "ProcessWire is a free and open source content management system (CMS), content management framework (CMF).",
     "headers": {
       "X-Powered-By": "ProcessWire CMS"
     },
+    "html": [
```
nurbek91, line 2511: I think this option is not suitable for determining the technology there will be a lot of  false-positive results

nurbek91 (changes requested):

src/technologies/p.json
seanhamlin: Will look for another sniff in the mean time, will close this.

PR #6839, opened 2022-10-12
Add Google Font API detection method

ysm-dev:
When using Google Fonts with the Next.js framework.
Next.js automatically optimizes the google font API, so wappalyzer cannot detect it.

(e.g. https://umami.is)

+ I'm not sure this line is correct.
`"<style data-href=\"https://fonts.googleapis.com/css*\""`


AliasIO:
Thank! Update the `html` patterns to `dom`.

PR #6953, opened 2022-11-02
Update Hyvä Themes

wigman:
The logo for Hyvä Themes is currently in use is outdated and the technology was not classified under eCommerce.

(I am the founder of Hyvä myself)


src/technologies/h.json
```diff
@@ -1534,9 +1534,9 @@
   },
   "Hyva Themes": {
     "cats": [
-      19
+      6
```
nurbek91, line 1537: I'm not sure about categories. We usually try to add only platforms to the ecommerce category. There are requirements that only one technology be identified under this category. For WordPress and Shopify we have separate WordPress/Shopify theme categories and for Magento, we do not have a separate category That's why I specified the category as 19

wigman: Hi @nurbek91! Wow, you are fast! It's admirable. Thanks for reviewing and merging this so quickly.

Our situation is actually a bit different from "just" a frontend team, as we've built our own frontend implementation where others build there themes on top of.
So, in the Magento/eCommerce ecosystem, we are competing with a lot of headless frontends. Magento's newest frontend is "PWA Studio" which is headless. They _are_ listed as a platform:
image

We've outgrown PWA Studio in 2 years time.
Our only other main competitor is Vue Storefront, again headless frontend for Magento. Which is also recognized as a platform.

Being recognized as a eCommerce frontend/platform is super important for us, to keep "competing" in graphs such as these:
image

You can see from that image that we are really making an impact on the eCommerce market. Getting recognized as a platform makes *a lot of difference* for us.

I hope you grant us that recognition. I'm happy to discuss/explain further in a 1-on-1 conversation, if that is useful.

peterjaap: Agreed with @wigman , Hyva is very much a platform and not *just a theme*.

nurbek91: Hey @wigman , It must have been my mistake then.
1. In this case, Vue StoreFront and PWA Studio should also be written to category 19
2. Create a separate Magento Headless category
3. Return Hyvä Themes  to category 6
Here we need the creator of the wappalyzer, he will tell you how to do it right.
Hey @AliasIO , need help!

wigman: Oh, that'd be super helpful. If you want, I could write identifiers for the traditional Magento frontend too (that one's called the Luma Theme). That way we can truly distinguish all frontends and compare.
It's probably a topic that applies to all the other eCommerce frameworks too. Shopify alone has a couple, if I'm not mistaken 😅.
And some are backend agnostic, like VueStoreFront.

nurbek91: Thanks Elbert, Hey @wigman , added new category #6962 and moved there  Hyvä Themes,  VueStoreFront, PWA studio.
Also, you can create an issue with a list of technologies that fall into this category, if possible, I will try to add them

wigman: @nurbek91 perfect! Thank you so much!

PR #6954, opened 2022-11-02
Template update

praetorian-thendrickson:
The Okta template was not detecting a lot of Okta instances. I updated the template to include current Javascript elements for better matching

### DIFF
--- a/src/technologies/o.json
+++ b/src/technologies/o.json
@@ -342,6 +342,9 @@
     "description": "Okta is a platform in the Identity-as-a-Service (IDaaS) category. Okta features include Provisioning, Single Sign-On (SSO), Active Directory (AD) and LDAP integration, the centralized de-provisioning of users, multi-factor authentication (MFA), mobile identity management.",
     "icon": "Okta.svg",
     "js": {
+      "OktaAuth": "",
+      "isOktaEnabled": "",
+      "oktaCurrentSessionUrl": "",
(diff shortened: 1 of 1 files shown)


src/technologies/o.json
```diff
@@ -342,9 +342,8 @@
     "description": "Okta is a platform in the Identity-as-a-Service (IDaaS) category. Okta features include Provisioning, Single Sign-On (SSO), Active Directory (AD) and LDAP integration, the centralized de-provisioning of users, multi-factor authentication (MFA), mobile identity management.",
     "icon": "Okta.svg",
     "js": {
-      "OktaAuth": "",
-      "isOktaEnabled": "",
-      "oktaCurrentSessionUrl": ""
+      "okta.cdnUrlHostname": "",
+      "okta.locale": ""
```
AliasIO, line 346: Should we just keep all five?

praetorian-thendrickson: I didn't see those on the recent Okta login pages I checked, ie:
```
$ curl https://praetorianlabs.okta.com  -s | grep -i -e 'oktaCurrentSessionUrl' -e 'isOktaEnabled' -e 'OktaAuth'
$ 
```
They are probably worth keeping if they match on other Okta technology pages 👍

AliasIO: I'll add them back in. 👍

PR #6956, opened 2022-11-02
Add Recharts

ysm-dev:
https://recharts.org/

- https://recharts.org/
- https://recharts.org/en-US/examples


src/technologies/r.json
```diff
@@ -705,6 +705,18 @@
...
     "website": "https://rechargepayments.com"
   },
+  "Recharts": {
+    "cats": [
+      25
+    ],
+    "description": "Recharts is a Redefined chart library built with React and D3.",
+    "icon": "Recharts.png",
+    "dom": [
+      "[class^=recharts-]"
```
nurbek91, line 715: Maybe replace with this 
examples:
https://coinloop.io/
https://coinmetro.com/
https://www.uvocorp.com/
```
"dom": {
      "div > div > svg.recharts-surface": {
        "attributes": {
          "class": ""
        }
      }
    },
```

Hey @AliasIO ,  Does this design help with broad matches and does it improve performance?

src/technologies/r.json
```diff
@@ -705,6 +705,18 @@
     ],
     "website": "https://rechargepayments.com"
   },
+  "Recharts": {
+    "cats": [
+      25
+    ],
+    "description": "Recharts is a Redefined chart library built with React and D3.",
```
nurbek91, line 712: ` "description": "Recharts is a component-based charting library, which is exclusively built for React applications."`

`"implies": "React",` ?

src/technologies/r.json
```diff
@@ -705,6 +705,18 @@
...
+    "icon": "Recharts.png",
```
nurbek91, line 713: I drew two versions of the icon. One is more readable but doesn't match the actual icon. In general, there is no such icon there, just text. Large text is not very readable. Generally choose what suits  best.
Recharts1
Recharts

nurbek91 (changes requested):
Hey mate, thanks for PR!!!

src/technologies/r.json
AliasIO: Yes it does, although the original is ok. As long as the number of elements that match the pattern is limited.

src/technologies/r.json
ysm-dev: [3c1e744](https://github.com/wappalyzer/wappalyzer/pull/6956/commits/3c1e744ee53f74da981b400ded14d06d8195c084)
[f54528c](https://github.com/wappalyzer/wappalyzer/pull/6956/commits/f54528cbbf2195c0223c46bda8da0820b42a56bf)

src/technologies/r.json
ysm-dev: [aff59ee](https://github.com/wappalyzer/wappalyzer/pull/6956/commits/aff59ee85cc75fb63eed5b97d669fc52bc4ba834)

nurbek91 (approved):
Thanks mate!!

PR #6958, opened 2022-11-02
I don't want my application to be detected anymore

simonvdfr:
I am the creator of Translucide


AliasIO:
Any particular reason why? :)

simonvdfr:
I no longer want the information around the CMS to be centralized and identifiable (also for security reasons).

AliasIO:
Removing the app from Wappalyzer doesn't make it any less identifiable or more secure.

PR #6959, opened 2022-11-03
Escape special regular expression characters

thc202:
Escape `{` to allow the regular expression to be used as is in Java as well.
Also escape `.`, for consistency with the first CSS regular expression.


nurbek91 (approved):
Hey mate, thanks for PR!! My mistake

PR #6967, opened 2022-11-08
Add magento frontends

wigman:
This adds the standard Magento 1 and Magento 2 frontends to the new "eCommerce frontends" category.

### DIFF
--- a/src/technologies/l.json
+++ b/src/technologies/l.json
@@ -1685,7 +1685,7 @@
     ],
     "oss": true,
     "scriptSrc": [
-      "requirejs-config\\.js",
+      "requirejs-config\\.js"
     ],
     "website": "https://github.com/magento/magento2"
   },
--- a/src/technologies/l.json
+++ b/src/technologies/l.json
@@ -1678,12 +1678,11 @@
     },
     "icon": "Magento.svg",
     "implies": [
-      "Magento\\;version:2"
+      "Magento\\;version:2",
+      "RequireJS",
+      "Knockout.js",
+      "jQuery"
     ],
-    "js": {
-      "Mage": "",
-      "VarienForm": ""
-    },
     "oss": true,
     "scriptSrc": [
       "requirejs-config\\.js",
--- a/src/technologies/l.json
+++ b/src/technologies/l.json
@@ -1659,6 +1659,37 @@
     },
     "website": "https://www.luigisbox.com"
   },
+  "Luma Theme": {
+    "cats": [
+      108
+    ],
+    "description": "Luma is the standard Magento 2 / Adobe Commerce frontend template.",
+    "cookies": {
+      "X-Magento-Vary": "",
+      "mage-cache-storage": "",
+      "mage-cache-storage-section-invalidation": "",
+      "mage-translation-file-version": "",
+      "mage-translation-storage": ""
+    },
+    "dom": {
+      "script[type='text/x-magento-init']": {
+        "exists": ""
+      }
+    },
+    "icon": "Magento.svg",
+    "implies": [
+      "Magento\\;version:2"
+    ],
+    "js": {
+      "Mage": "",
+      "VarienForm": ""
+    },
+    "oss": true,
+    "scriptSrc": [
+      "requirejs-config\\.js",
+    ],
+    "website": "https://github.com/magento/magento2"
+  },
   "Luna": {
     "cats": [
       105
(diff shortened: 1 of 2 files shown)


src/technologies/l.json
```diff
@@ -1659,6 +1659,36 @@
...
+    "icon": "Magento.svg",
+    "implies": [
+      "Magento\\;version:2",
+      "RequireJS",
+      "Knockout.js",
+      "jQuery"
+    ],
+    "oss": true,
+    "scriptSrc": [
+      "requirejs-config\\.js"
```
nurbek91, line 1688: It looks like there will be a lot of false-positive results. It's probably best to remove it.

src/technologies/m.json
```diff
@@ -30,6 +30,30 @@
...
+      "Prototype",
+      "jQuery"
+    ],
+    "js": {
+      "Mage": "",
+      "VarienForm": ""
+    },
+    "oss": true,
+    "scriptSrc": [
+      "skin/frontend/(?:default|(enterprise))"
```
nurbek91, line 53: It also looks like there will be false-positive results.

nurbek91 (commented):
Hey mate thanks for PR!!!

src/technologies/l.json
wigman: I got a bit lost in understanding how the regex works here, so I simplified it. But if you can help me with this one we can make it stricter :)

A full url to this file would look like:
`static/version1666692600/frontend/Magento/luma/nl_NL/requirejs-config.js`

So a general regex pattern would be:
`static\/version(\d+)?\/frontend\/(\w+)\/(\w+)\/(\w+)\/requirejs-config.js`

I saw you don't escape slashes, but do escape periods, so would it be this?
`static/version(\d+)?/frontend/(\w+)/(\w+)/(\w+)/requirejs-config\.js`

src/technologies/m.json
wigman: I haven't seen that pattern anywhere else to be honest. And you're using this literal url format to detect Magento 1 already :)

https://github.com/wappalyzer/wappalyzer/blob/master/src/technologies/m.json#L309

nurbek91: maybe this one ok https://publicwww.com/websites/%22skin%2Ffrontend%2Fdefault%22/

src/technologies/l.json
nurbek91: `static/version\d+/frontend/Magento/luma/[\w_]+/requirejs-config\\.js`

wigman: "Magento/Luma" will be different values in 99% of the times. Any agency called "MageShops", building a frontend for  "AcmeClothing" would create a theme witth the path:
`static/version1666692600/frontend/MageShops/AcmeClothing/en_GB/requirejs-config.js`
-or-
`static/version1666692600/frontend/AcmeClothing/default/en_GB/requirejs-config.js`
-or-
`static/version1666692600/frontend/AcmeClothing/uk/en_GB/requirejs-config.js`

So should we go with:
`static/version\d+/frontend/\d+/\d+/[\w_]+/requirejs-config\\.js` ?

nurbek91: > So should we go with:
> `static/version\d+/frontend/\w+/\w+/[\w_]+/requirejs-config\\.js` ?

Yes, this should help remove a significant amount of false positives. 
`static/version[\\d]{10}/frontend/\\w+/\\w+/[\\w_]+/requirejs-config\\.js`
can also be added to requires 
`"requires": "Magento",`
then only Magento sites will be checked. But headless Magento sites that we have not yet defined will not be detected, I think there are such sites, but such examples are probably less than 1 percent if I'm not mistaken

nurbek91: I think I'm a little confused here 
static/version1666692600/frontend/MageShops/AcmeClothing/en_GB/requirejs-config.js

src/technologies/m.json
nurbek91: In Magento itself, there is already a similar line, but there is 50% confidence

```

"scriptSrc": [
      "js/mage",
      "skin/frontend/(?:default|(enterprise))\\;version:\\1?1 (Enterprise):1 (Community)",
      "skin/frontend/\\;confidence:50\\;version:\\1",
      "static/_requirejs\\;confidence:50\\;version:2"
    ],
```
On this site, I see about 14 thousand results, probably on this data we can say that it looks like reality
https://publicwww.com/websites/%22skin%2Ffrontend%2Fdefault%22/

PR #6984, opened 2022-11-15
Added godaddy and arsys domain parking

dan-processbolt:
Finally getting back to issue 3796.

The first two added and working correctly are GoDaddy and Arsys.


AliasIO:
Thanks! PNG icons should be 32x32 px and locales need to be updated when adding a category, I'll update that.

PR #7008, opened 2022-11-21
Update Google maps

ysm-dev:
- Update icon 2500x2500 to 32x32
- add dom detection
  - https://www.embedgooglemap.net/
  - https://2022.jsconf.kr/en


AliasIO:
SVG dimensions don't really matter as larger ones don't take up more space.

PR #7013, opened 2022-11-21
Update strapi detection

ysm-dev:
- https://www.patriziagarganti.com/
image

### DIFF
--- a/src/technologies/s.json
+++ b/src/technologies/s.json
@@ -5471,7 +5471,8 @@
     ],
     "description": "Strapi is an open-source headless CMS used for building fast and easily manageable APIs written in JavaScript.",
     "headers": {
-      "X-Powered-By": "^Strapi"
+      "X-Powered-By": "^Strapi",
+      "x-powered-by": "^Strapi"
     },
     "icon": "Strapi.svg",
     "oss": true,


src/technologies/s.json
```diff
@@ -5471,7 +5471,8 @@
     ],
     "description": "Strapi is an open-source headless CMS used for building fast and easily manageable APIs written in JavaScript.",
     "headers": {
-      "X-Powered-By": "^Strapi"
+      "X-Powered-By": "^Strapi",
+      "x-powered-by": "^Strapi"
```
nurbek91, line 5475: hey mate, i think headers work case insensitive

raystinc: > hey mate, i think headers work case insensitive

but Strapi isn't detected on this site: https://www.patriziagarganti.com/

Do you know what is the reason?

ysm-dev: > hey mate, i think headers work case insensitive

but Strapi isn't detected on this site: https://www.patriziagarganti.com/

Do you know what is the reason?

nurbek91: yes mate, header parameter check only first file which at the top, other files we can check with only xhr parameter

ysm-dev: oh I see, didn't know that. Thanks!

PR #7015, opened 2022-11-21
Add Funding Choices to technologies

charisTheo:
Example pages using funding choices:

1. https://www.indiatoday.in/
2. https://www.phonearena.com/


AliasIO:
Thanks!

PR #7054, opened 2022-11-25
Update Cookiebot icon

ysm-dev:
- https://www.google.com/s2/favicons?domain=www.cookiebot.com&sz=256


AliasIO:
Thanks for all the PRs!

PR #7080, opened 2022-11-26
AOS Detection using HTML DOM

dhavalveera:
I have added the HTML DOM Element to detect the AOS which is used & installed using NPM instead of CDN Files.

### DIFF
--- a/src/technologies/a.json
+++ b/src/technologies/a.json
@@ -209,7 +209,7 @@
       "AOS.refresh": "",
       "AOS.refreshHard": ""
     },
-    "dom": ["[class*='aos-init']", "[class*='aos-animate']"],
+   "dom": " body[data-aos-easing]",
     "oss": true,
     "scriptSrc": [
       "unpkg\\.com/aos@([\\d\\.]+)/dist/aos\\.js\\;version:\\1",
--- a/src/technologies/a.json
+++ b/src/technologies/a.json
@@ -209,6 +209,7 @@
       "AOS.refresh": "",
       "AOS.refreshHard": ""
     },
+    "dom": ["[class*='aos-init']", "[class*='aos-animate']"],
     "oss": true,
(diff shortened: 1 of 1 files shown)


src/technologies/a.json
```diff
@@ -209,6 +209,7 @@
       "AOS.refresh": "",
       "AOS.refreshHard": ""
     },
+    "dom": ["[class*='aos-init']", "[class*='aos-animate']"],
```
nurbek91, line 212: Looks like a wide search and resource intensive, it is desirable to specify a specific tag where it can be applied.

nurbek91 (commented):
Hey mate, thanks for the PR!!
You should provide examples of websites where this technology is not defined and that these updates solve the problem. Preferably from 3-4 links if you add a new technology or at least 1 if it's a bug fix

dhavalveera:
> Hey mate, thanks for the PR!! You should provide examples of websites where this technology is not defined and that these updates solve the problem. Preferably from 3-4 links if you add a new technology or at least 1 if it's a bug fix

https://muqqabla.com/ => on this website AOS has been used by installing using `npm` not by the AOS CDN, so by the above PR, in the Websites where AOS is being used using npm package, can be detected using the DOM.

src/technologies/a.json
nurbek91: > Looks like a wide search and resource intensive, it is desirable to specify a specific tag where it can be applied.
Thanks for https://muqqabla.com/ 
similar example:
https://www.iii.com/
Can we use this trace. The body tag will need to be checked once.
body[data-aos-easing]

dhavalveera: I've updated the branch, and instead of searching the entire DOM, using `body[data-aos-easing]` to detect the AOS.

nurbek91 (approved):

PR #7125, opened 2022-11-29
Add Tally Detection

ysm-dev:
https://tally.so

image


src/technologies/t.json
```diff
@@ -528,6 +528,27 @@
     "saas": true,
     "website": "https://tallentor.com"
   },
+  "Tally": {
```
nurbek91, line 531: https://calvarychapel.com/
https://alexanderspatisserie.com/
https://www.happytech.life/

I'm not sure if it's worth defining the technology on these sites.
Also, if you follow the link, the current code can determine whether the technology can be tested, for example, here
https://tally.so/r/meAZlm
Might need to change some code here to
"scriptSrc": "//tally\\.so/",

Maybe we’ll add the widget category 5, it has a large priority and the technology will be visible at the top

nurbek91 (commented):
examples:
https://lightyshare.com/
https://www.hitradio.hu/
https://www.empara.fr/
https://lestalentsdalphonse.com/

src/technologies/t.json
ysm-dev: @nurbek91 
I think it's good to separate two technology between "Embed(Widget) and "script (non-embed)".

e.g.
- "Figma" vs "Figma Embed" (https://www.figma.com/developers/embed)
- "Youtube" vs "Youtube Embed" (https://support.google.com/youtube/answer/171780?hl=en)
- "CodeSandbox" vs "CodeSandbox Embed" (https://codesandbox.io/docs/learn/sandboxes/embedding)

it can be easily detected by "iframe" or not.

So here, it's better to separate "Tally" and "Tally Widget or Embed".

What do you think?

nurbek91: Quite a lot of work will need to be done, previously there was a policy all in one heap.
My powers are not enough here, we need to call @AliasIO

nurbek91: I don't quite understand how you want to share.
In my examples, only:

1. scripts
2. link to https://tally.so/r/meAZlm

nurbek91: Previously, I added a lot of technologies especially in Appointment scheduling categories
There were many similar situations there.
For example, a site has a script or iframe and sometimes a link to a third party.
In such cases, I asked Elbert whether it is worth indicating that the site uses this technology if it is just a link to the Tally platform, for example. He then allowed me.

ysm-dev: Okay, then

I think
```json
"scriptSrc": "//tally\.so/",
"dom": "a[href*='//tally.so/r/']"
```
is enough...? 🤔

nurbek91: yes what you need.
As for your idea of ​​separations, Elbert will probably read it and say whether or not to do this

ysm-dev: Thanks! [395c835](https://github.com/wappalyzer/wappalyzer/pull/7125/commits/395c83571afb0c031dd0b1aaae3a0f82b3e64405)

nurbek91 (approved):

PR #7129, opened 2022-11-30
fix "Choices" regexp

AsaiKen:
`|css` matches all scriptSrc including "css". It is not correct. I fix it.


src/technologies/c.json
```diff
@@ -1361,7 +1361,7 @@
     "icon": "Choices.png",
     "oss": true,
     "scriptSrc": [
-      "choices\\.js/|@([\\d.]+)(?:/assets)?(?:/scripts)?(?:/styles)?(?:/dist)?/choices(?:\\.min)?\\.js|css\\;version:\\1"
```
nurbek91, line 1364: I'm afraid this **scriptSrc** attribute only checks **js** files.
Could you provide a link where this technology is not defined.
Apparently we need to add the **dom** attribute and check the **css** file

AsaiKen: This rule detects at "https://www.aeyesec.jp/" by scriptSrc "https://static.parastorage.com/services/communities-blog-ooi/1.546.0/PostListViewerWidgetNoCss.bundle.min.js"
It seems false-positive.

nurbek91: Strange, but I do not see this technology.
Could you delete the cached results and look again

AsaiKen: Is this procedure for removing the cache okay?
https://user-images.githubusercontent.com/12981948/204837007-43ce6984-8b92-403c-9bb0-8c7deab830b3.mp4

nurbek91: Ah, ok i will take a look

nurbek91: ###  Hey mate, can u replace code with this one.

```
 "js": {
      "Choices": ""
    },
  "scriptSrc": [
      "choices\\.js(?:@|/)?([\\d\\.]+)?.+choices\\.min\\.js\\;version:\\1",
      "/modules/choices/js/choices\\.js"
    ],
```

###  Add new detection examples: 
http://grapp.ir/
https://www.vashdosug.ru/
https://portal.ciee.org.br/
https://kvf.fo/
http://www.designindaba.com/

### remove false-positive example:
https://www.aeyesec.jp/

AsaiKen: oh, it works fine.

PR #7204, opened 2022-12-14
Vitejs Detection using html + dom

dhavalveera:
### I'll see how many results can be collected, maybe over time there will be suggestions on how to improve the current code

examples:

- [https://vitejs.dev/](https://vitejs.dev/)
- [https://vitepress.vuejs.org/](https://vitepress.vuejs.org/)
- [https://troisjs.github.io/](https://troisjs.github.io/)
- [https://sli.dev/](https://sli.dev/)
- [https://vitejsvitegrjbyd-l5s3--5173.local-credentialless.webcontainer.io/](https://vitejsvitegrjbyd-l5s3--5173.local-credentialless.webcontainer.io/)


nurbek91:
Hey mate, In all examples, the technology is already defined
html is used in rare cases, for example when you need to read a comment
https://vitejsvitegrjbyd-l5s3--5173.local-credentialless.webcontainer.io/ only it is not defined here, but it looks like a stub and not the site itself

PR #7212, opened 2022-12-16
Update LaunchDarkly icon

ysm-dev:
https://launchdarkly.com/


AliasIO:
I think we should use the black version for this. I also think it's possible to include both light and dark version by embedding CSS (`prefers-color-scheme`) in the SVG.

ysm-dev:
> I think we should use the black version for this. I also think it's possible to include both light and dark version by embedding CSS (`prefers-color-scheme`) in the SVG.

[123560f](https://github.com/wappalyzer/wappalyzer/pull/7212/commits/123560fe6df5d7c1204e844e015066705de07723)

AliasIO:
Thanks!

PR #7214, opened 2022-12-17
Improved Shopify Detection

dhavalveera:
### Issue #6061 

https://www.seedlipdrinks.com/en-gb/


src/technologies/s.json
```diff
@@ -2401,7 +2401,11 @@
       "sdks\\.shopifycdn\\.com",
       "cdn\\.shopify\\.com"
     ],
-    "scripts": "shopifyTag",
+    "scripts": [
+      "shopifyTag",
+      "shopifyDomainName",
+      "shopifyAccessToken"
+    ],
```
nurbek91, line 2408: image
It seems that the current code works and defines Shopify as a platform.
`**scripts**` should be a little resource intensive it should be added when there are a lot of examples and when the technology is not defined

nurbek91 (commented):
Hey mate, probably should close the PR

Issue #6083, opened 2022-04-10
Track 'Rely' BNPL


rockeynebhwani:
Vendor site - https://www.rely.sg/
Category - Buy now pay later / Payment processors
Example Site - https://missfit.com.sg/collections/marimekko-collection/products/adidas-x-marimekko-training-tank-tee-hc6853-white
Detection - Look for script 'rely-shopify.min.js'

rockeynebhwani:
Still needed

nurbek91:
![image](https://user-images.githubusercontent.com/3456008/180084300-e6190af6-1298-484e-9735-14aa8cd31645.png)
![image](https://user-images.githubusercontent.com/3456008/180084333-59f936c6-877a-42f6-a16b-b78cb86a5808.png)

nurbek91:
I see JS traces but I don't see on the front of the offer site. Is it by any chance shopify app?
![image](https://user-images.githubusercontent.com/3456008/180085909-cce3125d-0c8b-45a9-980a-b0ade9d92418.png)
![image](https://user-images.githubusercontent.com/3456008/180085936-f71b7bf4-d729-43b6-8758-482b7dea73bc.png)

nurbek91:
https://missfit.com.sg/collections/marimekko-collection/products/adidas-x-marimekko-training-tank-tee-hc6853-white
https://www.picketandrail.com/products/kuka-2537-fabric-sofa-2-seater-color-c-1075?variant=16921021972550

rockeynebhwani:
@nurbek91 - looks like Rely is now part of Pace - https://www.pymnts.com/acquisitions/2022/singapore-fintech-pace-acquires-bnpl-firm-rely/ and the website link is https://payrely.com/

We are already tracking Pace - https://www.wappalyzer.com/technologies/payment-processors/pace/ so maybe we can just update Pace

rockeynebhwani:
.

PR #7215, opened 2022-12-17
Trace Rely BNPL

dhavalveera:
### Issue #6083 

https://www.picketandrail.com/products/kuka-2537-fabric-sofa-2-seater-color-c-1075?variant=16921021972550

### DIFF
--- a/src/technologies/p.json
+++ b/src/technologies/p.json
@@ -181,7 +181,10 @@
       "pacePay": ""
     },
     "saas": true,
-    "scriptSrc": "pay\\.pacenow\\.co",
+    "scriptSrc": [
+      "pay\\.pacenow\\.co",
+      "/rely-shopify.min.js"
+    ],
     "website": "https://pacenow.co/"
   },
   "Packlink PRO": {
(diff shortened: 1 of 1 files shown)


src/technologies/p.json
```diff
@@ -181,7 +181,10 @@
       "pacePay": ""
     },
     "saas": true,
-    "scriptSrc": "pay\\.pacenow\\.co",
+    "scriptSrc": [
+      "pay\\.pacenow\\.co",
+      "/rely-shopify.min.js"
```
nurbek91, line 186: Let's add a js object instead of a script. It is less resource intensive. If in other examples the js object is not loaded and can only be detected via scriptSrc then you can leave
```
 "js": {
      "pacePay": "",
      "rely_shop_currency": "",
      "rely_month_installment": ""
    },
```

nurbek91 (changes requested):
Hey mate, thanks. need small updates

src/technologies/p.json
nurbek91, line 186: `"/rely-shopify.min.js" - > "/rely-shopify\\.min\\.js"`

dhavalveera:
@nurbek91 done all the improvements you stated above

nurbek91 (approved):
thanks!!

PR #7216, opened 2022-12-18
Capterra Trace added

dhavalveera:
### Technology Request

Capterra - Analytics

**Examples:**

https://www.regfox.com/
https://share.rothys.com/
https://www.rentmanager.com/
https://share.litter-robot.com/
https://humanity.tcpsoftware.com/


src/technologies/c.json
```diff
@@ -541,6 +541,15 @@
     "scriptSrc": "^https?://api\\.captchme\\.net/",
     "website": "http://captchme.com"
   },
+  "Capterra": {
+    "cats": [
+      10
```
nurbek91, line 546: Capterra is definitely not a website analytics tool. It's more of a review category.
Site analysts measure js files or some kind of embedded pixel

src/technologies/c.json
```diff
@@ -541,6 +541,15 @@
...
+    ],
+    "description": "Capterra is a web analytics service that tracks and reports website traffic",
+    "dom": "a[href*='.capterra.com/'][target='_blank']",
```
nurbek91, line 549: "a[href*='.capterra.com/'][target='_blank'] > img[alt*='captera']", something like this should be 
Then we can tell the website is using Capterra and G2 badges.

src/technologies/c.json
```diff
@@ -541,6 +541,15 @@
...
+    "icon": "capterra.png",
```
nurbek91, line 550: Capterra
Try to send svg icon.

src/technologies/c.json
```diff
@@ -541,6 +541,15 @@
     "scriptSrc": "^https?://api\\.captchme\\.net/",
     "website": "http://captchme.com"
   },
+  "Capterra": {
+    "cats": [
+      10
+    ],
+    "description": "Capterra is a web analytics service that tracks and reports website traffic",
```
nurbek91, line 548: `"description": "Capterra is a technology review platform that connects sellers of B2B software to potential buyers.",

 `

nurbek91 (changes requested):
Hey mate, thanks for PR. need improvements

src/technologies/c.json
dhavalveera: what other Category will be there? can you tell me that, if possible?

dhavalveera: it'll be both `10` & `90` or only `90`?

src/technologies/c.json
dhavalveera: can you tell me more which trace to use or should I use both `a[href*=] img=[alt*=]`?

nurbek91: If it's a badge, we should check if the badge is a link inside with an img tag
Try to find 10 or more examples and make a universal pattern for identifying a badge

src/technologies/c.json
nurbek91: 90 only

src/technologies/c.json
dhavalveera: I didn't get you for `badge`

and should I use only `img[alt*='capterra']`? @nurbek91

src/technologies/c.json
dhavalveera: done

src/technologies/c.json
nurbek91: You should use a pattern that will uniquely define only badges and not links in an article, for example, which lead to a capterra.

dhavalveera: I checked all the examples where they've link only, not anything else @nurbek91

nurbek91: Perhaps it will be difficult for you to understand at this stage. You can close the PR or re-read and work on it. I don't know how else to explain what is required.

PR #7267, opened 2022-12-30
Nextjs Version Detection Improved

dhavalveera:
[Website Link](https://dev.vendor.muqabala.app/)  = where the issue was detected and below is the screenshot of the version detection

image

As you can see above image the version is coming like `\1` which I'm getting in both localhost as well as with the above-shared URL and this PR will resolve this which is tested locally, by the same website in localhost as well as with the above-shared URL


LoryPelli:
https://www.pastery.net/
This seems to happen also on this website with bootstrap

LoryPelli:
Check also on https://www.twitch.tv/ for Apollo version

AliasIO:
Fixed in 08ef520d47fcfcaf1cdbed3346f85706888eb610, thanks!

PR #7286, opened 2023-01-04
Add detection for WordPress themes Twenty Twenty-Two and Twenty Twenty-Three

felixarntz:
This PR intends to add detection for the WordPress default themes Twenty Twenty-Two and Twenty Twenty-Three, which are currently missing from the list of WordPress default themes that Wappalyzer detects.

Before reviewing the code, please read the following concerns:
* Those two themes (as well as likely any WordPress default themes in the future) follow a new theme development paradigm which, while it has several benefits, makes it quite a bit harder to identify such a theme being used.
* Twenty Twenty-Two still comes with a stylesheet that identifies the theme, so this one is still relatively straightforward, however it already lacks all other criteria that Wappalyzer relies on to detect previous WordPress default themes. Twenty Twenty-Three however does not load any stylesheet that identifies the theme, so here the PR is currently not working.
* The only thing that both themes _do_ load which we may be able to detect is some web font files that are within the theme's directory (`wp-content/themes/twentytwentytwo/assets/fonts` and `wp-content/themes/twentytwentythree/assets/fonts` respectively). However I'm unsure whether/how we can detect that via Wappalyzer.
    * I haven't found any technology property related to web fonts. Potentially we can somehow use the `css` property to detect the web font references? That's what I tried, but I don't know how that property is supposed to work, so any advice would be much appreciated.
    * FYI in both of the themes the web font reference is within inline CSS (i.e. not in an external CSS file). Maybe that helps.

For example URLs using the two themes, see:
* https://wp-themes.com/twentytwentytwo/
* https://wp-themes.com/twentytwentythree/

### DIFF
--- a/src/technologies/t.json
+++ b/src/technologies/t.json
@@ -2936,6 +2936,9 @@
     "dom": {
       "style#webfonts-inline-css": {
         "text": "/wp-content/themes/twentytwentythree/assets/fonts/"
+      },
+      "style#wp-webfonts-inline-css": {
+        "text": "/wp-content/themes/twentytwentythree/assets/fonts/"
       }
     },
     "icon": "WordPress.svg",
@@ -2953,6 +2956,9 @@
     "dom": {
       "link#twentytwentytwo-style-css": {},
       "style#webfonts-inline-css": {
+        "text": "/wp-content/themes/twentytwentytwo/assets/fonts/"
+      },
+      "style#wp-webfonts-inline-css": {
         "text": "/wp-content/themes/twentytwentytwo/assets/fonts/"
       }
     },
--- a/src/technologies/t.json
+++ b/src/technologies/t.json
@@ -2932,8 +2932,12 @@
     "cats": [
       80
     ],
-    "css": "\\/wp-content\\/themes\\/twentytwentythree\\/assets\\/fonts\\/",
     "description": "Twenty Twenty-Three is the default WordPress theme for 2023.",
+    "dom": {
+      "style#webfonts-inline-css": {
+        "text": "/wp-content/themes/twentytwentythree/assets/fonts/"
+      }
+    },
     "icon": "WordPress.svg",
     "pricing": [
       "freemium"
@@ -2945,9 +2949,15 @@
     "cats": [
       80
     ],
-    "css": "\\/wp-content\\/themes\\/twentytwentytwo\\/assets\\/fonts\\/",
     "description": "Twenty Twenty-Two is the default WordPress theme for 2022.",
-    "dom": "link#twentytwentytwo-style-css",
+    "dom": [
+      "link#twentytwentytwo-style-css",
+      {
+        "style#webfonts-inline-css": {
+          "text": "/wp-content/themes/twentytwentytwo/assets/fonts/"
+        }
+      }
+    ],
     "icon": "WordPress.svg",
     "pricing": [
       "freemium"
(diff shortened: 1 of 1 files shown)


felixarntz:
@AliasIO Do you have any advice on the issue I outline above? Is there a way for Wappalyzer to inspect web fonts loaded by a page, or web fonts referenced in CSS?

src/technologies/t.json
```diff
@@ -2928,6 +2928,33 @@
     "scriptSrc": "/wp-content/themes/twentytwentyone/",
     "website": "https://wordpress.org/themes/twentytwentyone"
   },
+  "Twenty Twenty-Three": {
+    "cats": [
+      80
+    ],
+    "css": "\\/wp-content\\/themes\\/twentytwentythree\\/assets\\/fonts\\/",
```
felixarntz, line 2935: As outlined in the PR description, this is not working (making the Twenty Twenty-Three detection fail, since that's the only thing that can identify it).

src/technologies/t.json
```diff
@@ -2928,6 +2928,33 @@
...
+      "freemium"
+    ],
+    "requires": "WordPress",
+    "website": "https://wordpress.org/themes/twentytwentythree"
+  },
+  "Twenty Twenty-Two": {
+    "cats": [
+      80
+    ],
+    "css": "\\/wp-content\\/themes\\/twentytwentytwo\\/assets\\/fonts\\/",
```
felixarntz, line 2948: As outlined in the PR description, this is not working.

src/technologies/t.json
nurbek91: examples
https://opposition24.com/
http://wp-themes.com/
http://playr.org/
https://wp-a2z.org/
https://techkyuniverse.com/

nurbek91: Hey mate yes confirm. but i couldn't detect when test it locally 

```
"dom": {
      "style#wp-webfonts-inline-css": {
          "text": "/wp-content/themes/twentytwentythree/assets/fonts/"
      }
    }, 
```
this code help me and its less resource intensive because we specify style ID

src/technologies/t.json
nurbek91: examples:
https://apkxmod.org/
https://impactwrestling.com/
http://mazochina.com/
http://seiichiegawa.jp/
https://stanceworks.com/
https://sola.ai/

```
"dom": {
      "style#wp-webfonts-inline-css": {
          "text": "/wp-content/themes/twentytwentytwo/assets/fonts/"
      }
    }, 
```

nurbek91 (changes requested):
Hey mate thanks for PR!!

src/technologies/t.json
nurbek91: "dom": {
      "style#wp-webfonts-inline-css": {
          "text": "/wp-content/themes/twentytwentytwo/assets/fonts/"
      }
    },

src/technologies/t.json
nurbek91:
```
"dom": {
      "style#wp-webfonts-inline-css": {
          "text": "/wp-content/themes/twentytwentythree/assets/fonts/"
      }
    }, 
```

felixarntz:
@nurbek91 @AliasIO Thank you for the guidance! I've updated the PR with the `dom` approach, works like a charm now!

src/technologies/t.json
felixarntz: @nurbek91 After further review, it looks like depending on the version of WordPress core (not the theme itself), the ID here can be either `wp-webfonts-inline-css` or just `webfonts-inline-css`. So I've updated the PR to cater for both.

AliasIO:
Cheers!

PR #7299, opened 2023-01-09
Detect partytown using script tag dom check

adamdbradley:
Detect partytown usage by checking if `script[type*='text/partytown']` is in the dom.

https://partytown.builder.io/partytown-scripts

### DIFF
--- a/src/technologies/p.json
+++ b/src/technologies/p.json
@@ -521,8 +521,10 @@
     ],
     "description": "Partytown is a lazy-loaded library to help relocate resource intensive scripts into a web worker, and off of the main thread.",
     "icon": "Partytown.svg",
-    "js": {
-      "partytown.lib": ""
+    "dom": {
+      "script[type*='text/partytown']": {
+        "exists": ""
+      }
     },
     "oss": true,
     "website": "https://partytown.builder.io/"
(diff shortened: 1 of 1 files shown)


AliasIO:
Thanks!

src/technologies/p.json
```diff
@@ -521,8 +521,9 @@
     ],
     "description": "Partytown is a lazy-loaded library to help relocate resource intensive scripts into a web worker, and off of the main thread.",
     "icon": "Partytown.svg",
+    "dom": "script[type*='text/partytown']"
     "js": {
-      "partytown.lib": ""
+      "partytown": ""
```
steve8708, line 526: @AliasIO I just noticed that this got readded

`partytown` as a global variable often won't exist, even when people are actively using partytown. the `dom` check is much more accurate. 

are these treated as an `AND` (so a website must have the dom selector match _and_ the `partytown` global variable to be flagged as a site that uses partytown, or is this an `OR` so if you have either or it's flagged?

if `OR` then we're all good, if `AND` we'll still have false negative detection issues tho

AliasIO: It's `OR`. 👍

PR #7311, opened 2023-01-11
fixing missing comma

brandonscholet:
I'm pulling the json automatically via python in https://github.com/brandonscholet/wappybird and it is now broken due to invalid json. Maybe there is a way to do linting to check for valid json? A was incorrect a few days ago I think


AliasIO:
Duplicate of #7302 👍

AliasIO:
I suggest using the latest tag rather than pulling from master for stability.

brandonscholet:
Thanks. I changed to meet that. There was still an issue with the tally thing though


def update_technologies_from_latest():
	print("updating technologies")
	technologies_file = os.path.expanduser('~/.python-Wappalyzer/technologies.json')
	technologies = {}
	 
	#get release page
	latest_release = requests.get('https://api.github.com/repos/wappalyzer/wappalyzer/releases/latest').json()
	#get zip from url
	zip_url = requests.get(latest_release['zipball_url'])
	myzip = ZipFile(io.BytesIO(zip_url.content)) 

	#parse files
	for listed_file in myzip.namelist():
		#get all technology files
		if "src/technologies" in listed_file and ".json" in listed_file:
			#extract file into json
			tech_json_file=myzip.read(listed_file).decode('UTF-8')
			#the latest release has an error
			if "t.json" in listed_file:
				tech_json_file = tech_json_file.replace('tally.so/embed/\'\"','tally.so/embed/\']\"')
			tech_json = json.loads(tech_json_file)
			#add to full json
			technologies = {**technologies, **tech_json}
		if "src/categories.json" in listed_file:
			#extract categories into json
			categories = json.loads(myzip.read(listed_file).decode('UTF-8'))
		#merge into one object
	combined_object = {'categories': categories, 'technologies': technologies}

	#write to file
	with open(technologies_file, 'w', encoding='utf-8') as tfile:
	    tfile.write(json.dumps(combined_object))
	    tfile.flush()
	print("done!\n")

brandonscholet:
if you want any linting setup help, let me know

AliasIO:
We have linting ([`yarn run validate`](https://github.com/wappalyzer/wappalyzer/blob/master/bin/validate.js)) that runs before every build. It doesn't currently validate the query selectors, we could add that in.

brandonscholet:
talked to the guy from soupsieve and he provided this. it might be handy to throw in somewhere in the python implementation of wappalyzer

https://github.com/facelessuser/soupsieve/issues/258#issuecomment-1380565764

brandonscholet:
image

worked super well.

AliasIO:
There is no official Python implementation. Not sure if this can be done natively in Node.js but something like that would be good.

PR #7327, opened 2023-01-15
technologies: Split Hostinger hosting from Hostinger CDN

ton31337:
We have two products, let's split into separate categories.

Signed-off-by: Donatas Abraitis <donatas.abraitis@hostinger.com>


AliasIO:
Thanks!

PR #7349, opened 2023-01-19
Jooma Module Category added

dhavalveera:
### Technology Request & Category Added

**`Examples:`**

https://www.allencounty.us/
https://milspecmonkey.com/
https://mariostrains.com/
https://www.mysysadmintips.com/
https://arkofhopeforchildren.org/


dhavalveera:
@AliasIO  @nurbek91 , I've added an Category for `Joomla Modules`, but when I am using that particular Category, then in the Browser Extension, that Category is isn't shown, what can be the cause?

PR #7350, opened 2023-01-19
Add GraphQL XHR detection

ysm-dev:
https://graphql.org/

ex)
- https://www.loom.com/pricing
- https://www.apollographql.com/

### DIFF
--- a/src/technologies/g.json
+++ b/src/technologies/g.json
@@ -2069,6 +2069,7 @@
       "store-config": "graphqlMethod"
     },
     "oss": true,
+    "xhr": "/graphql$",
     "website": "https://graphql.org"
   },
   "Graphene": {


src/technologies/g.json
```diff
@@ -2069,6 +2069,7 @@
       "store-config": "graphqlMethod"
     },
     "oss": true,
+    "xhr": "/graphql$",
```
nurbek91, line 2072: I guess this won't work. 
description
Hostnames of XHR requests.

ysm-dev: @nurbek91 oh I see.
Then how can I detect GraphQL on this site? : https://www.loom.com/pricing
This site looks using GraphQL.

nurbek91: @ysm-dev  at this stage it is not possible
I don’t know why it was done this way, apparently there are some reasons

ysm-dev: Thank you! @nurbek91

PR #7402, opened 2023-02-03
Add detection for WPML WordPress plugin

strategio:
WPML is a very popular WordPress plugin to run multilingual site, more details on https://wpml.org/.

**Note:** There's already [WooCommerce Multilingual](https://github.com/wappalyzer/wappalyzer/blob/master/src/technologies/w.json#L1581) in the existing technologies. It was initially a glue plugin of WPML to run with WooCommerce, but since version 5.0 it also works standalone to offer multi-currency. So I didn't make WPML a requirement of WCML.


src/technologies/w.json
```diff
@@ -1929,6 +1929,28 @@
...
+    "meta": {
+      "generator": "WPML"
+    },
+    "scriptSrc": "wp-content/plugins/sitepress-multilingual-cms",
+    "cookies": {
+      "wp-wpml_current_language": ""
+    },
+    "pricing": [
+      "low",
+      "recurring"
```
nurbek91, line 1949: onetime instead recurring maybe?

src/technologies/w.json
```diff
@@ -1929,6 +1929,28 @@
...
+  "WPML": {
+    "cats": [
+      87,
+      89
+    ],
+    "description": "WPML plugin makes it possible to build and run fully multilingual WordPress sites.",
+    "icon": "WPML.svg",
+    "oss": true,
+    "meta": {
+      "generator": "WPML"
```
nurbek91, line 1941: `"generator": "^WPML\\sver\\:([\\d\\.]+)\\;version:\\1"`
examples:
https://www.avature.net/
https://www.tlscontact.com/
https://mixpanel.com/

src/technologies/w.json
```diff
@@ -1929,6 +1929,28 @@
     ],
     "website": "https://www.wowza.com/video/player"
   },
+  "WPML": {
+    "cats": [
+      87,
+      89
+    ],
+    "description": "WPML plugin makes it possible to build and run fully multilingual WordPress sites.",
+    "icon": "WPML.svg",
```
nurbek91, line 1938: WPML
32x32

src/technologies/w.json
```diff
@@ -1929,6 +1929,28 @@
...
+    "oss": true,
+    "meta": {
+      "generator": "WPML"
+    },
+    "scriptSrc": "wp-content/plugins/sitepress-multilingual-cms",
```
nurbek91, line 1943: `"scriptSrc": "/wp-content/plugins/sitepress-multilingual-cms/.+\\.js(?:\\?ver=(\\d+(?:\\.\\d+)+))?\\;version:\\1",`

nurbek91 (commented):
Hey mate, thanks for PR!!

src/technologies/w.json
strategio: I assume this is correct, WPML is sold as a 1 year licence automatically renewable.

src/technologies/w.json
strategio: Thanks @nurbek91, adjusted!

src/technologies/w.json
strategio: @nurbek91, I did some tests and it's not working (I mean we can't get the version number from that). Also tried other patterns, unsuccessfully.

The problem is that we have "template" scripts with their own version number (not related to the plugin's version).
We actually may have 1 WPML script loaded on the frontend, but it depends on the site settings.

I prefer not trying to get the version from this rule.

strategio:
@nurbek91, thanks for you review!

FYI, I think all rules are "confident", but some are more likely to be found. Indeed, the script and the cookie might be available but it also depends on the site setup (depends on 2 different settings).

The rule that might be the most efficient for detection is the meta generator.

Let me know if you have other comments.

src/technologies/w.json
nurbek91: ok mate

nurbek91 (approved):

nurbek91 (approved):

nurbek91:
> @nurbek91, thanks for you review!
> 
> FYI, I think all rules are "confident", but some are more likely to be found. Indeed, the script and the cookie might be available but it also depends on the site setup (depends on 2 different settings).
> 
> The rule that might be the most efficient for detection is the meta generator.
> 
> Let me know if you have other comments.

Hey mate, everything looks great. Thanks for PR!!

strategio:
Hey @nurbek91, will you merge it? Or do need anything else from me? Thanks!

nurbek91:
> Hey @nurbek91, will you merge it? Or do need anything else from me? Thanks!



> Hey @nurbek91, will you merge it? Or do need anything else from me? Thanks!

Elbert will merge it. Approximately at the beginning of the each week.

strategio:
Thanks guys!

PR #7405, opened 2023-02-06
Adding UniFi OS

Darkiros:
Adding UniFi OS version detection.

Exemple of website : 
https://212.83.41.101:8443/manage/account/login
https://178.238.148.149:8443/manage/account/login
https://13.68.128.84:8443/manage/account/login


src/technologies/u.json
```diff
@@ -65,6 +65,23 @@
...
+    "cats": [
+      28
+    ],
+    "description": "UniFi OS is the operating system for UniFi products, which provides a user interface.",
+    "dom": {
+      "div[class*='unifiAccount__appNameAndVersion']": {
+        "text": "([\\d+\\.]+)\\;version:\\1"
+      }
+    },
+    "implies": "AngularJS",
```
nurbek91, line 78: For performance, it may be better to replace with requires?

src/technologies/u.json
```diff
@@ -65,6 +65,23 @@
...
+  "UniFi OS": {
+    "cats": [
+      28
+    ],
+    "description": "UniFi OS is the operating system for UniFi products, which provides a user interface.",
+    "dom": {
+      "div[class*='unifiAccount__appNameAndVersion']": {
+        "text": "([\\d+\\.]+)\\;version:\\1"
+      }
+    },
```
nurbek91, line 77: These things seem to duplicate js. Maybe it's better to leave only js?
unifiConstant.VERSION

nurbek91 (commented):
Hey mate, thanks for PR. just leave comments at your discretion

src/technologies/u.json
Darkiros: Okay, i will change it for "require".

src/technologies/u.json
Darkiros: Yes, but it is not the same thing. I put several sources for the detection. 
I can remove the dom detection if needed.

nurbek91: I checked in the examples you provided and it works in all js. If there are examples where there is no js then of course you can leave

Darkiros: Yes I check on other websites and it works without dom. I remove the dom part.

nurbek91 (approved):

PR #7445, opened 2023-02-17
Improve Vue.js detection

osbre:
As you can see [in the official docs](https://v3-migration.vuejs.org/breaking-changes/mount-changes.html#_3-x-syntax), in Vue 3 after the app is rendered - it will have `data-v-app` attribute.


src/technologies/v.json
```diff
@@ -1136,7 +1136,10 @@
     ],
     "css": "\\.vue-notification-group",
     "description": "Vue.js is an open-source model–view–viewmodel JavaScript framework for building user interfaces and single-page applications.",
-    "dom": ".vue-app",
+    "dom": [
+      ".vue-app",
+      "[data-v-app]"
+    ],
     "html": "<[^>]+\\sdata-v(?:ue)?-",
```
nurbek91, line 1143: Hey mate
`[data-v-app] `duplicate  `"html": "<[^>]+\\sdata-v(?:ue)?-", ` if i am not mistaken we can make sure if you share links for tests

osbre: Oh, you are right. Sorry then.

PR #7382, opened 2023-01-25
Added BigDataCloud API

bexdeep:
BigDataCloud provides various paid and free-to-use geolocation APIs for websites and mobile apps.

PR #7475, opened 2023-02-28
Updating regex of Bigdatacloud

bexdeep:


nurbek91:
previous request https://github.com/wappalyzer/wappalyzer/pull/7382
Hey @bexdeep , I noticed that in the previous examples the technology is not defined, let's change the code to this, maybe it will help
`"xhr": "api\\.bigdatacloud\\.net/"`

the problem should not be in the  forward slash

bexdeep:
Thanks @nurbek91 . I will make the change.

AliasIO:
Thanks!

PR #7492, opened 2023-03-05
Update MkDocs / documentation

nurbek91:
### just add implies python
Basic understanding of Python is required to use MkDocs effectively. MkDocs is a static site generator that is built using the Python programming language, and it uses Python to execute many of its core functions.

In order to create and customize a documentation website using MkDocs, you will need to be able to create and edit Markdown files, which is a simple markup language used for formatting text. Additionally, you may need to modify the configuration file and use Python to add custom plugins or functionality to your site.

While some parts of MkDocs can be managed through a user interface or configuration files, understanding Python basics can be beneficial when working with more complex customizations or troubleshooting issues.
### examples:
https://developers.engineerforce.io/

### DIFF
--- a/src/technologies/m.json
+++ b/src/technologies/m.json
@@ -2105,6 +2105,7 @@
     ],
     "description": "MkDocs is a static site generator, used for building project documentation.",
     "icon": "mkdocs.png",
+    "implies": "Python",
     "meta": {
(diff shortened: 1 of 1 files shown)


baseplate-admin:
Same can be said for sphinx.

( can we update that too ? )

PR #7502, opened 2023-03-09
Add Google Forms detection

osbre:
Some websites use Google Forms as a "backend" solution, though it would be cool to keep track of that.


AliasIO:
Thanks!

Issue #7513, opened 2023-03-12
Track 'FunCaptcha'


John4266:
Vendor Site - https://www.arkoselabs.com/
Category - Security
Example Site - https://github.com/signup
Detection - HTTP request to `https://api.funcaptcha.com`

PR #7514, opened 2023-03-13
FunCaptcha Trace Added

dhavalveera:
#7513


src/technologies/f.json
```diff
@@ -2057,6 +2057,15 @@
     "scriptSrc": "\\.fullstory\\.com/",
     "website": "https://www.fullstory.com"
   },
+  "FunCaptcha": {
+    "cats": [
+      16
+    ],
+    "description": "The Arkose MatchKey challenge-response mechanism is the strongest CAPTCHA ever made, and is designed to meet modern threats head-on.",
```
nurbek91, line 2064: FunCaptcha is a type of CAPTCHA, which is a security measure used to protect websites and online services from spam, bots, and other forms of automated abuse.

nurbek91 (changes requested):

dhavalveera:
how to check whether the `xhr` is working or not? because whenever I try to add the `xhr` it never worked for me.

Issue #7520, opened 2023-03-14
The same domain gives different results using node version and wappalyzer web


elfeffe:
I'm using the latest version; I just installed it using node.

I can see Google Tag Manager here https://www.wappalyzer.com/lookup/codeanddeploy.com/

But not using wappalyzer https://www.codeanddeploy.com

Do you have any idea why?

PR #7521, opened 2023-03-14
Google Tag Manager Trace Improved

dhavalveera:
https://www.codeanddeploy.com/ => on this website it was used as `gtag/js` in scriptSrc whereas, in technologies, we were checking for `gtm.js`, this solve the problem #7520


nurbek91:
image
updates duplicate row for Google Analytics technologies 
Before that, this problem was raised, but then they decided that this line was not an imply GTM 
`
https://www.googletagmanager.com/gtag/js is a URL that points to a JavaScript file that is used for integrating Google Analytics and other tracking tools with websites. The file is called the Global Site Tag (gtag.js), and it is provided by Google Tag Manager.`

PR #7524, opened 2023-03-14
Improve Inertia.js detection

osbre:
Oftentimes `Vary` header gets included on the first request without `X-Inertia` header, and therefore can greatly improve the detection.

- https://app.notionmailer.com/login


AliasIO:
Thanks!

PR #7567, opened 2023-03-26
Add OpenSwoole detection

osbre:
https://github.com/openswoole/ext-openswoole


nurbek91 (approved):

nurbek91:
Hey @osbre , now PR has a web server category, we also have a web framework category. What do you think will be right?
```
OpenSwoole is primarily a networking communication framework for PHP, rather than a web server or web framework. While it can be used to build web applications and supports popular PHP web frameworks like Laravel and Symfony, it is not a full-fledged web framework in itself.

OpenSwoole is focused on providing a high-performance, asynchronous, and event-driven programming model for PHP developers, which can be used to build various types of network applications, including web servers, WebSocket servers, TCP/UDP servers, and more.

However, OpenSwoole does provide some features that are commonly found in web frameworks, such as routing, request/response handling, and middleware support. This makes it easier for developers to build web applications using OpenSwoole, but it is not designed to replace traditional web frameworks entirely.

So, to summarize, OpenSwoole is primarily a networking communication framework for PHP, but it can be used to build web applications and provides some features commonly found in web frameworks.
```

PR #7569, opened 2023-03-27
Add support for JavaScript variable detections

osbre:
This change has a big positive impact on many of the detections. I'll explain why.

Currently, Wappalyzer has support for detecting things that are only defined within the `window` scope, like this one:
```html
<html lang="en">
<head>
  <title>Test</title>
  <script>
    // or window["Ziggy"] =  ... 
    window.Ziggy = () => {
      console.log("Ziggy");
    };
  </script>
</head>
<body>
</body>
</html>
```

As you can see, this example uses `window` to define a new library. However, there are situations when this is not the case and the library is defined like this:

```html
<html lang="en">
<head>
  <title>Test</title>
  <script>
    const Ziggy = () => {
      console.log("Ziggy");
    };
  </script>
</head>
<body>
</body>
</html>
```

In such a case, the registered library is considered a variable and is not available in the `window` scope. Calling it as part of the `window` will not give any result:

Chrome Devtools console output showcasing window.Ziggy and Ziggy commands

**So what is the solution for not "missing out" on variables to detect stuff?**

Well, before this PR it was simply ignored. **This PR adds support for JavaScript variable detection while maintaining compatibility with the previous algorithm and not breaking anything.**

To check if the variable is defined, it uses [`new Function('js code')()`](https://javascript.info/new-function#:~:text=The%20syntax%20for%20creating%20a,with%20the%20arguments%20arg1...) syntax as a way of checking if the variable is defined and getting its value. It is considered to be a faster and safer alternative to the `eval` approach.

Looking forward to your feedback @AliasIO, hoping that together we can make even more libraries detectable on lots of new websites.

Thank you.


AliasIO:
Good find. Accessing variables on the `window` object works for `var` declarations but not `const` or `let`.

Unfortunately using `Function()` is considered unsafe eval and will throw errors when a website restricts `script-src` in their [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). It doesn't look like the errors can be suppressed (you can test it here on github.com):
 
Screenshot 2023-03-28 at 10 59 45 am

I also don't think this will pass validation on the extension stores for security reasons.

I did implement a condensed version of your code in the Node.js version where we don't have these limitations, see 9bc44d8b6324adbc19bcbe8f9abd0e7b0c8fa0c6.

osbre:
Thanks for bringing it into the Node.js version! Indeed, while the `EvalError` can be handled with try-catch, the extension stores would not pass on the `Function()` usage.

AliasIO:
The error I shared was _with_ a try/catch. If we could handle it gracefully we could check if CSP is being used and fall back to the `window` method if needed. We'd still have the store validation issue though.

PR #7601, opened 2023-04-09
Change Fly.io category

osbre:
I think "PaaS" is more suitable rather than a "web server"


nurbek91 (approved):
Yes, Fly.io is a platform-as-a-service (PaaS) offering. It provides a platform for deploying, managing, and scaling containerized applications without the need for infrastructure management. With Fly.io, developers can focus on building their applications and leave the infrastructure management to the platform.

PR #7610, opened 2023-04-10
Update Odoo / cms

nurbek91:
fixed category: 
### example:
https://www.ossfactory.com/web/login
But I'm not sure about the categories anyway. Also not sure about the list of technologies in the implies array

Odoo is not primarily a Content Management System (CMS) in the traditional sense, but it does have some features that could be considered CMS-like. Odoo is an open-source enterprise resource planning (ERP) software that provides a suite of business applications that can be used to manage various aspects of a business, such as accounting, inventory, project management, and more.

In terms of CMS-like features, Odoo includes a website builder that allows users to create and manage website content using a drag-and-drop interface. This website builder includes a range of customizable templates, content blocks, and widgets that can be used to create a variety of website layouts and designs.

In addition, Odoo includes features for managing and publishing blog content, as well as tools for creating and managing email campaigns. These features provide some basic CMS-like functionality, although Odoo's primary focus is on business management and ERP functionality rather than content management.

Overall, while Odoo does offer some CMS-like features, it is primarily designed as an ERP solution and may not be the best choice for organizations that require a full-featured CMS.


nurbek91:
odoo implies Node.js ?
Odoo is primarily built using the Python programming language and relies on the Python ecosystem for its backend server-side processing. While Odoo uses JavaScript for its frontend user interface, it does not require or rely on Node.js specifically.

That being said, Odoo can be extended and customized using a variety of programming languages and technologies, including JavaScript and Node.js. For example, developers can use JavaScript and Node.js to build custom frontend interfaces, connect with external APIs or services, or automate certain tasks within the Odoo system.

However, it's important to note that Node.js is not a required component of the Odoo platform, and most Odoo development can be done using Python and the built-in Odoo development framework.

others Seems right: 
"Python",
 "PostgreSQL",
  "Less"

PR #7639, opened 2023-04-24
Fix Sucuri - remove WordPress requirement

kyletaylored:
Sucuri works with multiple platforms, not just WordPress. Removing this requirement opens up other sites that are also using Sucuri.

**Screenshot from their website:**
image


nurbek91:
Hey @kyletaylored , thanks!!
- In this case, we need to update the description:
"description": "Sucuri is a cybersecurity company that provides website security solutions and services. Its main focus is on protecting websites from malware, hacks, and other online threats.",

- fix category 
replace current 31 CDN, 87 WP plugin to 16 Security

- and let's replace the icon with svg:
Sucuri

nurbek91 (changes requested):
Need to update everything

kyletaylored:
Awesome, made those updates. I left CDN because it is still a CDN product, but removed WordPress plugin and added Security.

PR #7650, opened 2023-04-29
LottieFile Version Detection with JS Object

dhavalveera:
### Improvement

using `js: {}` object, detecting the Version of LottieFiles


nurbek91 (approved):
examples
https://pubwise.io/
https://www.ikwilindrukmaken.be/home

PR #7671, opened 2023-05-22
Update Google Publisher Tag Domains

gilbertococchi:
Many Origins are not classified as using "Google Publisher Tag" due to a buggy Google Publisher Tag Classification.

Default one: https://securepubads.g.doubleclick.net/tag/js/gpt.js

Limited Ads one: https://pagead2.googlesyndication.com/tag/js/gpt.js

Old: https://googletagservices.com/tag/js/gpt.js


AliasIO:
Thanks!

PR #7686, opened 2023-06-01
Update Plausible Analytics

osbre:
Plausible Analytics is built on top of the Phoenix Framework, which will also imply the usage of the Elixir programming language

https://github.com/plausible/analytics


osbre:
@nurbek91 Then I would suggest to exclude "implies" section for certain categories such as analytics.

PR #7690, opened 2023-06-02
Improve Drupal Multisite detection

kyletaylored:
This PR improves the Drupal Multisite detection (builds upon the work in #7687) to detect images in multisite paths and not just script files.


AliasIO:
Thanks!

PR #7736, opened 2023-06-11
fix: http protocol in websites should be https

meddulla:
Change http protocol to HTTPS in website links, eg. http://akamai.com to https://akamai.com


AliasIO:
Thanks!

PR #7757, opened 2023-06-13
taboola json changes

talivneh:
made changes in Taboola detection

### DIFF
--- a/src/technologies/t.json
+++ b/src/technologies/t.json
@@ -336,6 +336,7 @@
     "js": {
       "_taboola": "",
       "TRCImpl": "",
+      "TRCImpl.global": "",
       "TRC": "",
       "taboola_view_id": ""
     },
--- a/src/technologies/t.json
+++ b/src/technologies/t.json
@@ -335,9 +335,7 @@
     "icon": "Taboola.svg",
     "js": {
       "_taboola": "",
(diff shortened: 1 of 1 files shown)


src/technologies/t.json
```diff
@@ -335,9 +335,13 @@
     "icon": "Taboola.svg",
     "js": {
       "_taboola": "",
-      "_taboolaNetworkMode": "",
+      "TRCImpl": "",
+      "TRC": "",
```
nurbek91, line 339: Hey mate, thanks for finding a bug!!!
I would advise adding an additional check to reduce some of the possible false positives.
examples:
https://www.morhipo.com/
http://thestartmagazine.com/feed/summary
```
"js": {
  "_taboola": "",
  "_taboolaNetworkMode": "",
  "taboola_view_id": ""
  "TRC.cdsPips": "",
  "trc_item_ur"l: "//taboolanews\\.com/",
  "TRCImpl.global": ""
},
```

talivneh: added TRCImpl.global, the others doesn't necessarily exists.

talivneh: I removed TRCImpl  and TRC. There is no property that is necessarily on the TRC.

AliasIO:
Thanks!

PR #7765, opened 2023-06-14
Added technology: Microsoft .NET ClickOnce

TheTechromancer:
"ClickOnce is a deployment technology that enables you to create self-updating Windows-based applications that can be installed and run with minimal user interaction." (source: https://learn.microsoft.com/en-us/visualstudio/deployment/clickonce-security-and-deployment?view=vs-2022)

Google dork: `"Publisher" "Version" "ClickOnce and .NET Framework Resources"`

### DIFF
--- a/src/technologies/c.json
+++ b/src/technologies/c.json
@@ -4452,6 +4452,7 @@
         "text": "ClickOnce and .NET Framework Resources"
       }
     },
+    "css": "\\.ClickOnceInfoText",
     "website": "https://learn.microsoft.com/en-us/visualstudio/deployment/clickonce-security-and-deployment"
   }
 }
--- a/src/technologies/c.json
+++ b/src/technologies/c.json
@@ -4440,5 +4440,18 @@
     "oss": true,
     "scriptSrc": "(?:/([\\d\\.-]+))?/crypto-js(?:\\.min)?\\.js\\;version:\\1",
     "website": "https://github.com/brix/crypto-js"
+  },
+  "ClickOnce": {
+    "cats": [
+      47
+    ],
+    "description": "ClickOnce is a Microsoft .NET deployment technology that enables the creation of self-updating Windows-based applications that can be installed and run with minimal user interaction.",
+    "icon": "default.svg",
+    "dom": {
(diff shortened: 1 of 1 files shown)


src/technologies/c.json
```diff
@@ -4440,5 +4440,18 @@
...
     "website": "https://github.com/brix/crypto-js"
+  },
+  "ClickOnce": {
+    "cats": [
+      47
+    ],
+    "description": "ClickOnce is a Microsoft .NET deployment technology that enables the creation of self-updating Windows-based applications that can be installed and run with minimal user interaction.",
+    "icon": "default.svg",
+    "dom": {
+      "a[href='https://go.microsoft.com/fwlink/?LinkId=154571']": {
```
nurbek91, line 4451: Hey @TheTechromancer  this trail leads to the documentation.
It will not be correct if, by reference to the documentation, we determine that the website uses this technology

TheTechromancer: I agree that the link in isolation is not enough to fingerprint the technology. However, this is only one of two requirements. As seen directly below, the `<a>` tag must also contain exactly "ClickOnce and .NET Framework Resources`".

nurbek91: Ok, let's leave it to your experience. Do you have examples of websites?

TheTechromancer: Yes, here are a few:

* https://users.cs.northwestern.edu/~ian/Twig/SafeHomeBase/publish.htm
* https://dsm.deltek.com/DeltekSoftwareManager/
* http://www.wvdhhr.org/bph/hsc/support/install/birth/publish.htm

nurbek91 (approved):

src/technologies/c.json
nurbek91: ok thanks mate, its implies Microsoft .NET framework? then we can add 
`"implies": "Microsoft ASP.NET"`
I hope it passes further testing before merging

TheTechromancer: In this case I don't think it implies ASP.NET, since it's specifically for .NET desktop applications. E.g. it is also possible to host it on Apache, nginx, etc.

AliasIO:
Thanks!

Issue #7771, opened 2023-06-24
Track 'Colbass'


John4266:
Vendor Site - https://colbass.com/
Category - Widgets
Example Sites:
https://passportnews.co.il/article/183273
https://www.pc.co.il/advertisements/388530/
https://www.hidabroot.org/%D7%AA%D7%94%D7%99%D7%9C%D7%99%D7%9D-%D7%9C%D7%A8%D7%A4%D7%95%D7%90%D7%94
https://www.ias.co.il/%d7%a8%d7%90%d7%a9%d7%99/%d7%93%d7%9c%d7%aa%d7%90-%d7%90%d7%99%d7%99%d7%a8%d7%9c%d7%99%d7%99%d7%a0%d7%a1-%d7%a2%d7%95%d7%91%d7%a8%d7%aa-%d7%9c%d7%98%d7%99%d7%a1%d7%94-%d7%99%d7%95%d7%9e%d7%99%d7%aa-%d7%91%d7%a7%d7%95-%d7%90/
Detection: `audio#colbass-tts-audio`

PR #7667, opened 2023-05-15
Add Ultimate Bulletin Board / message-boards

nurbek91:
### website:
https://www.ubbcentral.com/purchase.php
### examples:
https://ubbdev.com/forums/ubbthreads.php/forum_summary.html
https://board.moparts.org/ubbthreads/ubbthreads.php
https://clubsi.com/

PR #7774, opened 2023-06-26
Colbass Trace Detection added

dhavalveera:
#7771


src/technologies/c.json
```diff
@@ -2516,6 +2516,18 @@
     "scriptSrc": "https?://www\\.hashing\\.win/scripts/min\\.js",
     "website": "https://www.coinimp.com"
   },
+  "Colbass": {
+    "cats": [
+      5
+    ],
+    "description": "Colbass is AI System that transforms text into real-human voiceovers.",
+    "dom": "audio#colbass-tts-audio",
```
nurbek91, line 2524: Please check whether it is possible to determine the technology using this trace. The less resource intensive option should be
```
"js": {
      "colbassRunning": ""
    },
```

src/technologies/c.json
```diff
@@ -2516,6 +2516,18 @@
...
+    "icon": "colbass.png",
```
nurbek91, line 2525: 32x32
Colbass
Icon with transparent background

src/technologies/c.json
```diff
@@ -2516,6 +2516,18 @@
...
+    "pricing": [
+      "low"
+    ],
```
nurbek91, line 2528:
```
"pricing": [
      "low",
      "onetime",
      "recurring"
    ],
```

src/technologies/c.json
nurbek91: https://github.com/wappalyzer/wappalyzer/pull/7667#issuecomment-1557672416
Here, too, an icon with a transparent background is needed, if not, could you write that the request cannot be completed

nurbek91 (changes requested):
Hey mate, thanks for PR!!

src/technologies/c.json
dhavalveera: I didn't get you for the 2nd message where you've mentioned another PR

nurbek91: ok never-mind

nurbek91 (approved):

PR #7778, opened 2023-06-29
Add Moveable library detection

osbre:


src/technologies/m.json
```diff
@@ -2743,6 +2743,15 @@
     },
     "website": "http://movabletype.org"
   },
+  "Moveable": {
+    "cats": [
+      59
+    ],
+    "description": "Moveable is a JavaScript library that makes elements movable, scalable, rotatable, draggable and resizable.",
+    "dom": ".moveable-control-box",
```
nurbek91, line 2751: Hey mate, 
Do you have examples of websites where the technology is used.
Is it possible to detect via JS attribute moveable.min.js I saw this line but I can't find examples current detection option is resource intensive

PR #7779, opened 2023-06-29
Add WayForPay detection

osbre:
Examples:
- https://wayforpay.com/en/demo-shop


src/technologies/w.json
```diff
@@ -620,6 +620,16 @@
     "requires": "WordPress",
     "website": "https://music.flatfull.com/waveme/about/"
   },
+  "Wayforpay": {
+    "cats": [
+      41
+    ],
+    "description": "Wayforpay is a payment processing service provider based in Europe.",
+    "dom": "form[action*='secure.wayforpay.com']",
```
nurbek91, line 628: instead **dom** and **scriptSrc** maybe only js will be enough? 
```
"js": {
      "WayforpayInitParams": ""
    },
```

nurbek91 (changes requested):
Hey mate, thanks!!
All edits as an alternative that is responsible for optimisation

src/technologies/w.json
osbre: No, we cannot rely only on JS because there are cases when HTML form is used (I believe such usage is also very active, possibly even more popular than the JS library)

nurbek91: Fine. Do you have examples other than demo. I could not find examples where all these traces are used.

PR #7780, opened 2023-06-29
Replace `Spatie` logo with the SVG version

osbre:


nurbek91 (approved):
Hey mate, thanks!!!
why does it look big even though its size is 33x33?
Could send size 32x32?
Spatie

osbre:
@nurbek91 Updated the icon. I think it's just GitHub displays it like that. If you open it anywhere else - it will be 32x32. (not the big version)

nurbek91 (approved):

PR #7783, opened 2023-06-30
fix: Lightbox selector should have balanced quotes

kingthorin:


kingthorin:
Duplicate, disregard.

PR #7793, opened 2023-07-07
Moat update

max-ostapenko:
Moat is still alive, but was acquired by Oracle long ago


nurbek91 (approved):
Hey mate, thanks for PR!!

PR #7801, opened 2023-07-09
Flarum implies Laravel usage

osbre:
Flarum uses Laravel and is built on top of it. If the programming language (PHP) is being detected, it would make sense for the framework to be detected as well.


ceopeo (changes requested):
Hey mate, thanks!!!!
Apparently the technology has not been updated for a long time. Let's make some more improvements

PR #7854, opened 2023-07-19
Add Progress WS_FTP / miscellaneous

francisco-sec:
Added support for Progress WS_FTP, using dom fingerprint.

Website:
https://www.progress.com/ws_ftp

Examples:
https://ftp.atitesting.com/ (Recent Web Transfer Client)
https://sftp.nlacrc.org/thinclient/ (Older Web Transfer Client)


src/technologies/p.json
```diff
@@ -3025,6 +3025,16 @@
     ],
     "website": "https://www.profitwell.com/"
   },
+  "Progress WS_FTP": {
```
ceopeo, line 3028: Can we add such a line? This would help optimise the search.
"requires": "Microsoft ASP.NET",

src/technologies/p.json
```diff
@@ -3025,6 +3025,16 @@
...
+    "cats": [
+      19
+    ],
+    "cpe":"cpe:2.3:a:progress:ws_ftp_server:*:*:*:*:*:*:*:*",
+    "description": "The Progress WS_FTP software provides both an effective, highly manageable FTP server and the ability to provide each user a secure FTP client, assuring both reliability and security for data sharing processes.",
```
ceopeo, line 3033: Progress WS_FTP is a file transfer client software developed by Progress Software Corporation, supporting FTP, FTPS, SFTP, and HTTPS protocols with features like drag-and-drop support, file synchronization, and encrypted data transmission.

src/technologies/p.json
```diff
@@ -3025,6 +3025,16 @@
...
+    "dom": "spinner[class='spinner-wrap'][percent=''], form[name='formLogin'][action='login.aspx' i][id='formLogin']",
```
ceopeo, line 3034: Looks like a very broad search. Are there other indirect signs that will add confidence to us?
`"requires": .....`

ceopeo (changes requested):
Hey mate, thanks!!!

src/technologies/p.json
francisco-sec: Sure, that would be a great improvement. Just checked in several instances, and wappalyzer was able to recognize ASP.NET in all of them

src/technologies/p.json
francisco-sec: done

src/technologies/p.json
francisco-sec: Indeed, this one was the best I could found after analyzing dozens of different instances.
WS_FTP recent instances are using angular and bootstrap for the frontend...
However, if we add those requires, we might start having false negatives on old instances

It took me a long time to find this dom pattern, but please find below some others (only valid for recent instances):
Title: "Web Transfer Client" (Could get a lot of false positives)
Using url: /ThinClient/WTM/public/index.html
Having 'percent=\"{{progress}}\"' on html

Let me know your thoughts, and if only detecting recent instances would be a better tradeoff

ceopeo: ` "scriptSrc": "/ThinClient/(?:WTM|WebResource)(?:\\.axd|/public)",`

ceopeo: Can we replace dom with scriptSrc. After adding "requires": "Microsoft ASP.NET", the code is much better optimized and won't run every time.

If you are not sure  **dom** gives 100 confidence you can change like this
"dom": "form[name='formLogin'][action='login.aspx' i][id='formLogin']\\;confidence:40",
"scriptSrc": "/ThinClient/(?:WTM|WebResource)(?:\\.axd|/public)\\;confidence:60"",

francisco-sec: It was a great idea using that scriptSrc. Unfortunately, we can not rely only on scriptSrc as it will lead to a false negative rate of 30%.
Combining the dom and the ScriptSrc we can reach a false negative rate of 10%, also, both solutions won't produce false positives.
Taking this numbers into account, do you agree in keeping dom and scriptSrc?

ceopeo: Below 50 percent, the extension does not show in the results, but the npm driver will show at least 0, 30, 60
maybe something like this 
```
"dom": "spinner[class='spinner-wrap'][percent=''], form[name='formLogin'][action='login.aspx' i][id='formLogin']\\;confidence:30",
"scriptSrc": "/ThinClient/(?:WTM|WebResource)(?:\\.axd|/public)\\;confidence:30"",
```

francisco-sec: Dom filter like that is able to detect almost all cases, so does it make sense to give it a confidence of 30?

ceopeo: it's just that there are no keywords similar to the names of technologies, so that there were no false positive results, I wanted to include additional parameters

francisco-sec: I see, it makes sense to me now. Thanks for all your help!

PR #7863, opened 2023-07-24
Read from optional custom technologies file

rudiedirkx:
I have secret technologies I want to detect, so I need a custom JSON file somewhere. This way wappalyzer tries to read `wappalyzer-custom-technologies.json` **from the cwd**, so the user decides where to run it with what custom file.

It'd be even nicer if it was a cli option, but the technologies are global, so I don't see how.


AliasIO:
You can use [wappalyzer-core](https://www.npmjs.com/package/wappalyzer-core) to achieve this with minimal scripting.

rudiedirkx:
I'd have to recreate all of `cli.js`, `driver.js` and `wappalyzer.js`, no? And copy all the technologies from this repo. That's not minimal scripting IMO. Adding custom config to this repo is less scripting.

PR #7884, opened 2023-07-28
fix cloudfront incorrect implies

meddulla:
Just because a frontend loads, eg. a javascript file, hosted in Cloudfront does not imply the backend/website is using AWS as a PAAS.


AliasIO:
The Cloudfront detection doesn't trigger on assets like JavaScript files, only if the page itself it served by it. In that case it should list AWS as well.

meddulla:
That is correct, had a bug on my end.

PR #7906, opened 2023-08-08
chore: update Nuxt logo and website

Atinux:
As well as detecting Nuxt 3 website (even though it should already work) by adding `useNuxtApp` detection


AliasIO:
Thank you!

